PR #3, opened 2015-01-13
Renaming file to react naming conventions

scherler:
Hi to use your great work we need to rename the file otherwise our loader is complaining. Could you please apply my pull request, TIA

### DIFF
--- a/index.jsx
+++ b/index.jsx
@@ -1,0 +1,75 @@
+/**
+ * @jsx React.DOM
+ */
+
+var React = require('react');
+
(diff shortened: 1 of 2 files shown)


paramaggarwal:
We might need to add `extensions` config to `browserify` config in `package.json` for this change to also work with Browserify. Will verify.

joscha:
This is still an issue. Version `1.0.1` brings a `.js` ending instead of a `.jsx` one again.

PR #10, opened 2015-02-15
Added several improvements

Pomax:
- removed pragma (React changed to advocating using .jsx instead, a while ago)
- dropzone now ignores the hardcoded styling when an explicit classname is passed in
- minor tweak for onDrop, to return an array of files, rather than the incredibly obtuse FileList

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,3 @@
-/**
- * @jsx React.DOM
- */
-
 var React = require('react');
 
 var Dropzone = React.createClass({
@@ -47,6 +43,7 @@ var Dropzone = React.createClass({
     }
 
     if (this.props.onDrop) {
+      files = Array.prototype.slice.call(files);
       this.props.onDrop(files);
     }
   },
@@ -57,7 +54,7 @@ var Dropzone = React.createClass({
 
   render: function() {
 
-    var className = 'dropzone';
+    var className = this.props.className || 'dropzone';
     if (this.state.isDragActive) {
       className += ' active';
     };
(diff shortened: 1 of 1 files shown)


paramaggarwal:
Good ones. I'll change the file name to `.jsx` again, as was recommended in one of the other PRs.

index.js
```diff
@@ -67,9 +64,13 @@ var Dropzone = React.createClass({
       height: this.props.size || 100,
       borderStyle: this.state.isDragActive ? "solid" : "dashed"
     };
- 
+
+    if (this.props.className) {
```
paramaggarwal: This one looks like a problem to me. It means only if one gives a `className` prop, will it pick the `style` prop.

index.js
```diff
@@ -67,9 +64,13 @@ var Dropzone = React.createClass({
...
+      style = this.props.style;
+    }
+
```
paramaggarwal: This one looks like a problem to me. It means only if one gives a `className` prop, will it pick the passed-in `style` prop.

Pomax: good point, let me amend that

Pomax: updated so that style is always props.style OR the hardcoded default, however, if a className is explicitly passed in, the styling *has* to come either from the passed in style, or from plain CSS/LESS/SASS/Stylus/whatever.

Pomax:
any word on what you'd like with this PR? (I'd very much like to see it merged, the baked-in styling despite specifying my own className in 1.0.0 is still a big problem)

paramaggarwal:
Oops, I had missed it. Merged in, thanks!

paramaggarwal:
Published to npm as v1.0.1 (https://github.com/paramaggarwal/react-dropzone/releases/tag/v1.0.1)

Pomax:
super, thanks so much!

PR #13, opened 2015-04-01
Add previews to file objects

aleemb:
Add the preview property to file object which contains the client-side preview of the just uploaded file. This can come in handy to show thumbnails of just uploaded files.


paramaggarwal:
Good idea, I'll update README with this feature and push to npm.

PR #22, opened 2015-05-13
custom onDragOver and onDragLeave can be inserted via props.

timothyjlaurent:


timothyjlaurent:
Looks like this somehow breaks the onDrop functionality.

PR #23, opened 2015-05-13
Allow custom onDragOver and onDragLeave to be injected as props

timothyjlaurent:
This now works. Potentially, instead of adding to onDragOver, we could make an onDragEnter which would have the same effect (I think) as this PR.


timothyjlaurent:
Argggh.... Dragging onto child elements causes the dragLeave event to fire.

PR #24, opened 2015-05-15
Turn off onClick support by default

jackfranklin:
Just used this plugin (which is awesome BTW!) on a project where we wanted drag and drop support but not clicking, and made this change to support that. It means you have to pass in `supportClick={true}` to turn the click support. What do you think? Be happy to change the approach but probably quite useful to be able to customise this stuff?

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,12 @@
 var React = require('react');
 
 var Dropzone = React.createClass({
+  getDefaultProps: function() {
+    return {
+      supportClick: true
+    };
+  },
+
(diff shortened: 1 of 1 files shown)


paramaggarwal:
Oh good idea, but usually on websites where I am able to drag and drop, I am also able to click - so I believe the default should be `supportClick={true}`.

jackfranklin:
:+1: Change incoming!

jackfranklin:
@paramaggarwal all set :)

PR #29, opened 2015-05-21
Support setting multiple to false to enable single file drop/pick

matiaslarsson:
Example: <Dropzone multiple={false} onDrop={this.onDrop} size={150}></Dropzone>

Currently just takes the first file in the dropped array if the user drops multiple files. (I don't think it's good to disallow dropping if the user tries to drop several files on the drop zone.)


paramaggarwal:
Like!

PR #30, opened 2015-05-21
adding a file input name attribute, necessary for some file upload plugins

ishumadan91:
Please see if this seems appropriate to you. Just tried using with some other file upload plugin and it wasn't working because `name` attribute in the input tag was not present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ var DropzoneDemo = React.createClass({
     render: function () {
       return (
           <div>
-            <Dropzone onDrop={this.onDrop} size={150} >
+            <Dropzone onDrop={this.onDrop} size={150} inputName='myFileUpload' >
               <div>Try dropping some files here, or click to select files to upload.</div>
             </Dropzone>
           </div>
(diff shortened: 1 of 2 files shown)


paramaggarwal:
How about if we expose a React `ref` as a property on `<Dropzone>`. So `.getFileInput()` on the component would return `this.ref.fileInput`. Fits in with React's method of keeping a reference and does not need any configuration.

ishumadan91:
We can do that but I would still suggest putting an inherent name field. Even if we don't make it configurable, and tell about it in the documentation will be fine. name attribute may be very much required and we should provide ease to those using the plugin. Also it complies with the html standards to have a name in an input field

nickhudkins:
:+1:

paramaggarwal:
Please correct me if I am wrong, but having a `name` field that other libraries can reference is only useful for the default browser `input` element and not the drag and drop behaviour right? This library is primarily for drag and drop use. Using a default `input` field does not require the use of this library?

nickhudkins:
If this was purely used for asynchronous file upload, then you are correct, because onDrop is called, and all the file handling can be done there. However, if you'd like to simply POST the form traditionally, assuming that on the backend you'd like to know what files are what, a named input is needed :)

What COULD be done is another file input could be added, and the value could be set onDrop, but that seems awfully hacky.

If you take a look at this codepen:
http://codepen.io/nickhudkins/pen/MwPrPJ

you'll see the file field has no name, which if you post some data, and take a look at:
http://requestb.in/nvzb0mnv

you will see the file data is not coming through in a manner than could be used.

Go ahead and fork the codepen, add a name field to the file field and you will see it works wonderfully.

paramaggarwal:
@nickhudkins Yes, the example you give is perfectly valid when default `input` is used. But this library is uses HTML5 `onDrop` events which do not work as far as I know without normal form post uploads. The only reason there is an `input` element in this library is actually to support the scenario of a click and then a dropdown file selector.

The library won't work the other way round where you drag and drop a file and it becomes set on the input element and hence will not work with a simple browser form post action.

nickhudkins:
Because the component does support clicking, we should avoid breaking the simple browser form post.

As far as drop goes, you are absolutely correct.

paramaggarwal:
Makes sense guys, thanks for pushing for this. Good to merge once the conflicts are resolved.

paramaggarwal:
Please don't rely directly on the `input` field, it only serves a secondary purpose to the primary drag and drop feature. If you need a more advanced `input` field, please directly create your own.

PR #33, opened 2015-05-26
add styleFunc prop, shouldComponentUpdate

awli:
I was using this for my own purposes and made a couple of changes:
1. I added a `styleFunc` prop that takes a function which takes `props` and `state`. The custom style prop isn't too useful now as the style is static, and you can't style based on `isDragActive`.
2. I added a `shouldComponentUpdate`, because previously when you drag over the area it updates at a very fast frequency because of the onDrag.

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,6 +18,7 @@ var Dropzone = React.createClass({
     onDrop: React.PropTypes.func.isRequired,
     size: React.PropTypes.number,
     style: React.PropTypes.object,
+    styleFunc: React.PropTypes.func,
     supportClick: React.PropTypes.bool,
     accept: React.PropTypes.string,
     multiple: React.PropTypes.bool
@@ -69,6 +70,10 @@ var Dropzone = React.createClass({
     }
   },
 
+  shouldComponentUpdate: function(nextProps, nextState) {
+    return this.state.isDragActive != nextState.isDragActive;
+  },
+
   open: function() {
       this.refs.fileInput.getDOMNode().click();
   },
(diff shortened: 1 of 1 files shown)


paramaggarwal:
1. Ideally the `isDragActive` state needs to be exposed out to parent with `onDragStart` on `onDragEnd` events of something similar. Then the `style` prop would be changed accordingly. This should be cleaner.
2. Agree that the component updates continuously while onDrag is in progress. But relying just on `this.state.isDragActive != nextState.isDragActive` is prone to too many bugs. Going to wait for a better solution.

awli:
1. I'm not sure how much of a good idea exposing events to parents is, as this isn't a very React-like thing to do (i.e. have the data flow up to the parent). An alternative perhaps is to have a `style` and `styleActive` prop, but that might result in a large amount of duplication (for example, if the component is heavily styled but the change in the active state is only one or two properties).
2. That's fair, I guess that the update is not very expensive anyway as the DOM probably doesn't update regardless. An alternative is to modify the `onDrag` method to only call `this.setState` if the state has actually changed?

paramaggarwal:
1. In the PR, the parents gets access to the `props` and `state` object as parameters to the function sent as a `prop` - that doesn't sound right either. Sending separate styling for `active` and `idle` states might make more sense, especially if we could send an array to the `style` object and have the `active` style override `idle` style.
2. Yes, the correct thing to do would be to `setState` selectively.

cassus:
:+1: for styling separate styling of base / active state. Either by using styleFunc or with style + styleActive.
Even styleFunc is not that bad as the parent object already knows about the porps, that is public API. If you make state.isActive also part of the public and stable API than I see no problem with using styleFunc.

But then the styleFunc signature can be simplified to styleFunc({isActive}) as the owner already knows about the props they are giving so that can be in the closure for styleFunc and the only exposed state would be isActive. This would also make the public API surface explicit.

paramaggarwal:
Added support for `activeStyle` in `v2.0.1`.

PR #36, opened 2015-06-05
Update transformer

Martindelataille:


paramaggarwal:
You'll need to update the same in `package.json` also.

PR #41, opened 2015-06-22
Clear out file input before click to allow user to select same file again

brendon-gonzalez:
Inputs with a type of `file` store the file path in their value and if you select the same file again, you need to clear out the value so that the correct drop events can retrigger. Ran into this issue when a user was adding a file, deleting it via our file manager and then re-adding that file afterwards. This simply clears the input value before attempting to click it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -70,7 +70,9 @@ var Dropzone = React.createClass({
   },
 
   open: function() {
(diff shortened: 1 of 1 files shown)


paramaggarwal:
Thanks! Published to npm as `v1.2.2`.

PR #44, opened 2015-07-06
Adding option to make dropzone transparent

Sapphire64:
I wanted to use dropzone component as an transparent wrapper with high-level files API around my components, in this usage case I don't really need to have that styles and rewriting styles option with dummy data looks wrong to me.

What do you think about having this option?

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,7 +4,8 @@ var Dropzone = React.createClass({
   getDefaultProps: function() {
     return {
       supportClick: true,
-      multiple: true
+      multiple: true,
+      transparent: false
     };
   },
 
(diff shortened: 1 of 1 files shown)


paramaggarwal:
Just saying `style={{}}` seems fine.

Sapphire64:
@paramaggarwal But to know that it makes any effect you need to check sources. Maybe we should at least mention it somewhere?

paramaggarwal:
Very true! I'll do that once I find some time. :)

Sapphire64:
Ok, closing this one :)

paramaggarwal:
No no, leave it open as an open task to complete!

paramaggarwal:
Resolved via https://github.com/paramaggarwal/react-dropzone/pull/46

PR #46, opened 2015-07-20
Optionally disable style object in a clearly explicit manner and control size better with `width` and `height` props.

robmclarty:
I added an explicit `noDefaultStyles` prop to make it clear what's happening when someone doesn't want to use the default `style` object (for example, when using a custom CSS class). 

I found using `style={{}}` cryptic and unclear for anyone else looking at the code, not to mention that it wasn't obvious that I needed to pass in an empty object to get my CSS working properly in the first place (see https://github.com/paramaggarwal/react-dropzone/issues/42).

I also added separate `width` and `height` props for better, basic, control over the size of the Dropzone element (not everyone wants a square). But I left the original `size` prop for backwards compatibility and as an easy alternative if you *do* want a square.

I reformatted the render return value onto multiple lines to make it easier to read.

Finally, I added additional documentation to the README to cover these new changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,11 +12,11 @@ Usage
 
 Simply `require('react-dropzone')` and specify an `onDrop` method that accepts an array of dropped files. You can customize the content of the Dropzone by specifying children to the component.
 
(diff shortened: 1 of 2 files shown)


robmclarty:
I've removed the `noDefaultStyles` prop in preference of detecting the existence of either the `style` or `className` props to decide if the default styles should be applied.

paramaggarwal:
This is great, thanks!

PR #47, opened 2015-07-25
Hidden input[type=file] should have name attribute

nickhudkins:
This will allow react-dropzone to work onClick with standard browser POSTs. addresses conflicts in #30

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,7 +4,8 @@ var Dropzone = React.createClass({
   getDefaultProps: function() {
     return {
       supportClick: true,
-      multiple: true
+      multiple: true,
+      inputName: 'fileInput'
     };
   },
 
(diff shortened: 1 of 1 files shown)


paramaggarwal:
@nickhudkins But the component does not have a <form> wrapper around the <input> which means that simple browser form post anyways does not work.

nickhudkins:
True... so... I guess I am confused what you intended the use of this library to be. 

As I see it it is an enhanced file input field, which could be used in a <form> and in that case, the name attribute is important.

I guess my largest question for you is, what is the disadvantage of treating this as a semantic file input at it's core, and the additional functionality being just that, additional.

paramaggarwal:
@nickhudkins The library is intended to be used as a drag and drop component. There is no clean way to make this work as a simple `input` field. I'll see if the dropped file can be set as the file on the `input` field itself. As if it was selected from the dropdown - then it would be pretty cool.

paramaggarwal:
Closing as the primary feature of the library is drag and drop, and the `input` field is only to allow selection by clicking. To use with regular form POSTs, please directly use your own `input` field.

PR #48, opened 2015-07-29
Disable onDragOver activation if effect is not all

voronianski:
When dropping file into browser `effectAllowed` will be `"all"` that's why we can check this in order to skip setting active state on dropzone when it's not necessary, see demo:


index.js
```diff
@@ -42,6 +42,14 @@ var Dropzone = React.createClass({
     e.stopPropagation();
     e.dataTransfer.dropEffect = 'copy';
 
+    var effectAllowed = e.dataTransfer.effectAllowed;
+
+    // set active drag state only when file is dragged into
+    // (in mozilla when file is dragged effect is "uninitialized")
+    if (effectAllowed !== 'all' && effectAllowed !== 'uninitialized') {
+      return;
```
paramaggarwal: Please don't `return` here, as we need the `this.props.onDragOver(e)` to still be called always. Instead wrap ``` this.setState({
       isDragActive: true
     });``` inside an `if (effectAllowed === 'all' || effectAllowed === 'uninitialized')`

voronianski: @paramaggarwal updated, thanks!

voronianski:
@paramaggarwal updated

paramaggarwal:
I didn't even know about such an API. Thanks!

voronianski:
@paramaggarwal :wink: thanks!

voronianski:
@paramaggarwal when do you plan to bump npm version?

paramaggarwal:
Done. `v1.2.4`.

PR #51, opened 2015-07-31
Adding a display name to avoid the component to display as Unknown

ialex:
Adding a display name to avoid the component to display as Unknown in React Developer Tools

image


paramaggarwal:
Good one. This broke when JSX was removed from the library.

PR #52, opened 2015-08-02
Allow width / height to accept % as strings

prakhar1989:
Hi Param,

Thanks for this react component. :grinning: 

I wanted to provide `width:100%` to the dropzone but it gave a `propType` validation error. Here's a quick fix that allows the width and height of the dropzone to be set in %age (as well as pixels) without giving the `failed Proptype` error.

:beers:

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,8 +22,14 @@ var Dropzone = React.createClass({
     onDragOver: React.PropTypes.func,
     onDragLeave: React.PropTypes.func,
     size: React.PropTypes.number,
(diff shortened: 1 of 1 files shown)


paramaggarwal:
Please set via `style` instead: [index.js#L113](https://github.com/paramaggarwal/react-dropzone/blob/master/index.js#L113).

prakhar1989:
Ah yes! Totally forgot about that. Thanks for the quick turnaround. :+1:

PR #54, opened 2015-08-05
Use default styles and merge in user set ones

ravloony:
Hi,

I propose a patch to allow users to set one or more styles, without losing all the other defaults.

This is because I want to have the dropzone have an automatic width, by setting {width: '100%'} without losing the dashed border which becomes solid on dragging a file.

My code is a little verbose, but should work anywhere.

Thanks.

### DIFF
--- a/index.js
+++ b/index.js
@@ -102,15 +102,24 @@ var Dropzone = React.createClass({
       className += ' active';
     }
 
+    var defaultStyle = {
+      width: this.props.width || this.props.size || 100,
+      height: this.props.height || this.props.size || 100,
+      borderStyle: this.state.isDragActive ? 'solid' : 'dashed'
+    };
     var style = {};
     if (this.props.style) { // user-defined inline styles take priority
       style = this.props.style;
-    } else if (!this.props.className) { // if no class or inline styles defined, use defaults
-      style = {
-        width: this.props.width || this.props.size || 100,
-        height: this.props.height || this.props.size || 100,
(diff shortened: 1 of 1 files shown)


paramaggarwal:
How about expose `this.state.isDragActive` instead in some way?

ravloony:
Thanks for the reply!

How about a dragActiveStyle prop which would apply only when this.state.isDragActive is true?

paramaggarwal:
Too complex, simplest pattern is of the component to send stuff up to the parent via callbacks and then the parent can send updated props like `style` to the component.

paramaggarwal:
Please use the `onDragOver` and `onDragLeave` prop function that you can pass - this is how the border setting is done internally - https://github.com/paramaggarwal/react-dropzone/blob/master/index.js#L33-L60

ravloony:
Isn't that two way data flow? I'd say it's cleaner for the component to manage it's own state, rather than asking the parent to update a prop on an event. If the parent is always going to update the style when the drag happens, then why not put that logic in the component itself?

paramaggarwal:
My main concern is that this change brings more switches and controls to an otherwise simple component. Hence against the idea of adding more `props`.

Also the React pattern is to have stateless components, pass data down using props, pass events up using callbacks and maintain as much state as possible at the very top level.

ravloony:
Ok, fair enough! Thanks for your time!

paramaggarwal:
Thanks for understanding. Also with this PR, there would not have been a way to override the default styling then.

PR #55, opened 2015-08-11
allow passing in custom activeClassName

tikotzky:
### Issue
Right now `react-dropzone` adds a class of `active` while dragging a file over the component. The issue is that due to the global nature of css if you have an `.active` style defined in any other stylesheets it's styles will get applied to the `react-dropzone` instance.

### Solution
Allow passing in the `activeClassName` as a prop this way if you can choose a globally unique classname (by using something like css modules) for the active state.

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,7 +27,8 @@ var Dropzone = React.createClass({
     style: React.PropTypes.object,
     supportClick: React.PropTypes.bool,
     accept: React.PropTypes.string,
-    multiple: React.PropTypes.bool
+    multiple: React.PropTypes.bool,
+    activeClassName: React.PropTypes.string
   },
 
   onDragLeave: function(e) {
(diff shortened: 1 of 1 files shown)


tikotzky:
for reference in order to get this behavior right now you need to do something like
```
	getInitialState() {
		return { isActive: false };
	},

	handleDragOver() {
		this.setState({ isActive: true });
	},

	handleDragLeave() {
		this.setState({ isActive: false });
	},

	handleDrop() {
		this.setState({ isActive: false });
	},

	render() {
		return (
			<Dropzone
				onDrop={this.handleDrop}
				onDragOver={this.handleDragOver}
				onDragLeave={this.handleDragLeave}
				className={this.state.isActive ? 'uniqueClassName' : '')}>
		);
	}
```

this pr would allow you to instead do
```
	render() {
		return <Dropzone activeClassName='uniqueClassName'}>;
	}
```

paramaggarwal:
Thanks!

tikotzky:
Thank You!

Will you be publishing a new version to npm soon?

PR #56, opened 2015-08-12
moved main component from div to button to work with mobile

redenz:


paramaggarwal:
I tried this out and it did not work on mobile Safari for me.

redenz:
Try this one - I noticed the same thing the other day.  I changed it back to a div but enabled TouchEvents and used the onStartTouch to trigger it.  Appears to be working

paramaggarwal:
I just realised - this library has no use on a mobile as there is no concept of drag and drop of files on mobile. To support file inputs, one can simply use a normal `input` tag. For this reason, I'll be closing this PR.

PR #64, opened 2015-09-07
make input a self closing tag and move position of children

conorhastings:
in react 0.14, a warning is thrown when you pass children to a tag that HTML standards say is self closing. 
screen shot 2015-09-07 at 12 15 20 pm

### DIFF
--- a/index.js
+++ b/index.js
@@ -142,15 +142,14 @@ var Dropzone = React.createClass({
         onDragLeave={this.onDragLeave}
         onDrop={this.onDrop}
       >
+        {this.props.children}
         <input
           type='file'
           ref='fileInput'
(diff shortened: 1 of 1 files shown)


paramaggarwal:
Ah, good catch. Thanks for the fix!

paramaggarwal:
Oh, there is a problem now. Previously, the click area would immediately become the entire element. Now this `input` will probably be a very tiny spot. Will need an `onClick` in code now...

conorhastings:
I haven't had that issue, i've still been able to click the whole div, and open the file viewer, maybe i'm not sure what you mean.

paramaggarwal:
Yes, I misunderstood. It works, ignore.

PR #65, opened 2015-09-09
Implemented accept attribute support for dropzone.

### LINKED ISSUE #57: Accept Not working on file drop

okonet:
Introduced 2 new callbacks (`onDropAccepted` and `onDropRejected`).
Added `rejectClassName` prop to allow custom styling when drop will be rejected.
Closes #57

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 var React = require('react');
+var accept = require('attr-accept');
 
 var Dropzone = React.createClass({
 
@@ -16,7 +17,9 @@ var Dropzone = React.createClass({
   },
 
   propTypes: {
-    onDrop: React.PropTypes.func.isRequired,
+    onDrop: React.PropTypes.func,
+    onDropAccepted: React.PropTypes.func,
+    onDropRejected: React.PropTypes.func,
     onDragEnter: React.PropTypes.func,
     onDragLeave: React.PropTypes.func,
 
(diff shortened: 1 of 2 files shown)


paramaggarwal:
This is great, thanks!

index.js
```diff
@@ -24,16 +27,26 @@ var Dropzone = React.createClass({
...
     activeClassName: React.PropTypes.string,
+    rejectClassName: React.PropTypes.string,
 
     disableClick: React.PropTypes.bool,
     multiple: React.PropTypes.bool,
+    accept: React.PropTypes.string,
+  },
+
+  allFilesAccepted: function(files) {
+    return files.every(file => accept(file, this.props.accept))
```
paramaggarwal: is `.every` available on all browsers? I have always used `.map` instead.

okonet: Don't you use babel to compile the source? https://github.com/paramaggarwal/react-dropzone/blob/master/package.json#L7

The only thing is you'll probably need to add .babelrc with `"optional": ["runtime"],`

okonet: But anyway https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

okonet: BTW I just realized this will not work because of arrow function. Do you mind using ES6 in your project? I could submit another PR with the complete solution.

paramaggarwal: There is already a `babel` pipeline, so we can use anything.

okonet: Right, but not sure it will be transformed with current settings. Let me re-check and send a new PR.

quietcodes:
If rejectClassName is not defined, it is adding an "undefined" class, which seems like an oversight.

okonet:
Thanks. Seems like an issue to me. Mind filing it and I'll take a look soon.

PR #69, opened 2015-09-11
Use Chrome's ability to see dragged files without breaking other browsers

### LINKED ISSUE #68: Possible typo in onDragEnter
I'm getting errors on this line in the onDragEnter function:

```javascript
var dataTransferItems = Array.prototype.slice.call(e.dataTransfer ? e.dataTransfer.items : e.target.files);
```

because e.dataTransfer.items is undefined. I believe this is supposed to read ```e.dataTransfer.files``` to match up with the rest of the usages. 

For reference here's the mozilla documentation on the DataTransfer object: https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer

okonet:
Closes #68

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,8 +41,13 @@ var Dropzone = React.createClass({
   onDragEnter: function(e) {
     e.preventDefault();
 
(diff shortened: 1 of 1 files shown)


paramaggarwal:
Thanks!

PR #71, opened 2015-09-14
webpack config + allow for using 'activeStyle' independently of 'style'

burnsed:
This change allows for setting the 'activeStyle' independently of the 'style'. This is for cases where someone might want to use css classes to define the base style and but use inline-styles for the OnDragOver case.

Also included is a webpack config for building out a module that can be included in a <script> tag (instead of using `require`). 

Thanks!


paramaggarwal:
Thanks!

PR #73, opened 2015-10-01
Maintain drag-active state while hovering child elements.

marnusw:
The component works fine with text children, but when DOM elements or components are added as children the hover-active state is not maintained since the `onDragLeave` event fires when moving over children.

A solution is complicated by the fact that the order in which the `onDragEnter` and `onDragLeave` events fire on the top most element and its children is non-deterministic and depends on the speed with which the mouse moves over the various elements.

However, in the end, once you have left the parent the same number of elements entered must have been left as well, so counting the events on the way in and out turns out to be an effective way to solve the problem.

I'd love to hear your comments.

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,12 +34,19 @@ var Dropzone = React.createClass({
     accept: React.PropTypes.string,
   },
 
+  componentDidMount: function() {
+    this.enterCounter = 0;
+  },
+
   allFilesAccepted: function(files) {
     return files.every(file => accept(file, this.props.accept))
   },
 
   onDragEnter: function(e) {
     e.preventDefault();
+
+    // Count the dropzone and any children that are entered.
+    ++this.enterCounter;
 
     // This is tricky. During the drag even the dataTransfer.files is null
     // But Chrome implements some drag store, which is accesible via dataTransfer.items
@@ -66,6 +73,11 @@ var Dropzone = React.createClass({
   onDragLeave: function(e) {
     e.preventDefault();
 
+    // Only deactivate once the dropzone and all children was left.
+    if (--this.enterCounter > 0) {
+      return;
+    }
+
     this.setState({
       isDragActive: false,
       isDragReject: false
@@ -78,6 +90,9 @@ var Dropzone = React.createClass({
 
   onDrop: function(e) {
     e.preventDefault();
+
+    // Reset the counter along with the drag on a drop.
+    this.enterCounter = 0;
 
     this.setState({
       isDragActive: false,


okonet:
I've solved this by adding CSS `pointer-events: none` to the drag area.

marnusw:
Thanks for the tip @okonet! I'll check it out and close this if it solves my problem.

marnusw:
So @okonet's suggestion does indeed solve my problem. :smile: IE still does not support `pointer-events`, but at the moment that doesn't matter since it doesn't support File Drag&Drop either.

I think I will update this PR to include an addition to the docs to help others who might encounter this problem.

Or, we could just add this style rule automatically within the component; and if there might be some reason to not always do that allow disabling it with a prop. What do you think @paramaggarwal & @okonet?

okonet:
I'd add as a part of the base style to the component. Less hassle for developers. Don't think this can break something. I also think there should be no prop for that.

marnusw:
Hmm, there is a slight snag with this strategy. Disabling pointer events on the base style means the drag events never trigger on the dropzone. So, the children have to be wrapped in an additional `span` or other component with the `pointer-events: none` style applied to that wrapper. Then the drag triggers on the dropzone, but none of its children.

There is also a further unfortunate result: If you include a button as a child which can be clicked to open the file explorer, that button's pointer events are now disabled, meaning no pop-up messages and no clicks. The `pointer-events: auto` style has to be applied to the button to fix this. Not the end of the world, but also not ideal considering the `pointer-events: none` is hidden "magic" behavior.

And, now this leads to further problems: When dragging over the button those events are now bubbled to the dropzone who reacts to them again. This causes the hover active/inactive state of the dropzone to become messed up. e.g. in my case when I drag off the button the dropzone remains in-active; only when you completely drag off it and then back over it does it activate again.

Perhaps "counting events" per my initial could end up being more predictable?

(And possibly better supported... For instance if IE implements File API in the future, but still no `pointer-events`.)

okonet:
Yeah, that sounds reasonable to me.

Sincerely yours

Andrey Okonetchnikov 

@okonetchnikov 
http://okonet.ru

marnusw:
Thanks @okonet, I've reverted to the event counting implementation and it seems to work quite well in both Chrome and Firefox without any side-effects. If you approve of these changes, @paramaggarwal, I think this is ready to be merged.

index.js
```diff
@@ -34,6 +34,10 @@ var Dropzone = React.createClass({
     accept: React.PropTypes.string,
   },
 
+  componentDidMount: function() {
+    this._enterCnt = 0;
```
okonet: I would not save bytes here and go for explicit naming: `enterCounter` or similar. Let minifier take care of byte shaving...

marnusw:
Done, thanks.

okonet:
Closes #75

tony99nyr:
+1 @paramaggarwal Excited to give this a try.

paramaggarwal:
Thanks @marnusw @okonet

JonathanWi:
Hey I'm having an issue with the pointer-event strategy: I want to achieve a full-body dropzone that only display when the file is over the body of the page. I so have my dropzone in position fixed above all content, which causes the content below to not be clickable anymore.
I then apply the `pointer-events: none` to the container in order to be able to click below, with the caveat that the dropzone doesn't work from here.

What's the solution here @marnusw @okonet ?

Thanks a lot

okonet:
The more I think about it, the more I believe it should not be used for a "whole document is a drop zone" case. I think it might break a lot of stuff.

Instead, you probably should handle this on document level from your root component via `componentDidMount` method. 

I can imagine you could do only basic `dragOver` detection here and then render the dropzone component. WYT?

JonathanWi:
Thanks for your comment, sorry for getting back to you so late! 

> The more I think about it, the more I believe it should not be used for a "whole document is a drop zone" case. I think it might break a lot of stuff.

I think you summed up my issue here, I might not have chosen the right component for the job. For a while I switched to (React Dynamic Dropzone)[https://github.com/overthinker29/react-dynamic-dropzone], an unmaintained fork of your repo. What's interesting is that it tackles the issues mentioned above by generating the dropzone on drag over, so, while the listener is always here, you don't need any `pointer-events` hack. The downside being that — because the input file is generated on the fly — I lose access to the `open()` function, the click being unable to look for the `ref` that is not created.

Do you have any idea how I could fix this? I hope I'm being clear enough, and thanks a lot for your time!

okonet:
I think this is the only proper way: detect drag and mount the dropzone. I think this could be a special wrapper component, which one can attach anywhere in your component tree to provide such functionality _additionally_ to the current one. Something like `Dropzone.GlobalWrapper` or similar. This approach could also solve the `ref` I guess?

If you feel like it, please submit a PR.

PR #77, opened 2015-10-08
Use react-dom to reference DOM

pukapukan:
To suppress warning introduced by react v0.14

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 var React = require('react');
+var ReactDOM = require('react-dom');
 var accept = require('attr-accept');
 
 var Dropzone = React.createClass({
@@ -116,7 +117,7 @@ var Dropzone = React.createClass({
   },
 
   open: function() {
-    var fileInput = React.findDOMNode(this.refs.fileInput);
+    var fileInput = ReactDOM.findDOMNode(this.refs.fileInput);
     fileInput.value = null;
     fileInput.click();
   },


romanenko:
@okonet Yes, you are right. But this requires react `>= 0.14`, and it means that there will be a need to release a new version with breaking changes.

okonet:
Well, I don't think you can do `require('react-dom')` with React 0.13, so this should be a major release either way. Or am I missing something?

pukapukan:
@okonet is right.  I was naively thinking react-dom would be compatible with previous versions of react.  I'll close this one as @okonet said - it'll be a major release to upgrade to react add dependency to v0.14.  Cheers

paramaggarwal:
Keeping this open to eventually support React 0.14.

index.js
```diff
@@ -116,7 +117,7 @@ var Dropzone = React.createClass({
   },
 
   open: function() {
-    var fileInput = React.findDOMNode(this.refs.fileInput);
+    var fileInput = ReactDOM.findDOMNode(this.refs.fileInput);
```
newoga: Do we actually need ReactDOM here?

We should test to see if `this.refs.fileInput` is enough... That way this project doesn't have a dependency on `ReactDOM`

newoga:
FYI, just let a comment on a line. `this.refs.fileInput` should actually return the DOM node in 0.14.0. This is described here: https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#dom-node-refs

index.js
okonet: Yes, I'm aware of that. Thanks!

newoga: Okay cool :)

okonet:
Can we have it rewritten without using ReactDOM and `findDOMNode`?

okonet:
I'll merge #84 if it's okay for everyone.

pukapukan:
Lovely go ahead please

okonet:
Ok cool. Closing this one in favor of #84

PR #84, opened 2015-10-24
Upgrade package for react-0.14.0, use findDOMNode

kumarharsh:
This PR addresses the points raised in #77

### DIFF
--- a/index.js
+++ b/index.js
@@ -131,7 +131,7 @@ var Dropzone = React.createClass({
   },
 
   open: function() {
-    var fileInput = React.findDOMNode(this.refs.fileInput);
+    var fileInput = this.refs.fileInput;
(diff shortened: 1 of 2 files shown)


okonet:
This looks good. Can you please remove version bump from it? I'll do it myself. Thanks!

kumarharsh:
Alright

kumarharsh:
Although there should be another commit on the 2.x branch, which changes the dep from

`"react": ">=0.13"`

to 
`"react": "^0.13"`

or something similar, because although the 2.x branch does work with 0.14 as of now, things may change later, and the `>=` does not offer any guarantee of compatibility whatsoever.

okonet:
Yeah. I can take care of it

PR #89, opened 2015-11-02
Decouple reject and active classnames from className

### LINKED ISSUE #88: State based classNames require className
For `activeClassName` or `rejectClassName` to work, `className` must be set.

jking90:
fixes #88


paramaggarwal:
:+1:

PR #90, opened 2015-11-02
Update index.js to add backward compatibility for react

ripitrust:
React ref are not point to the Native DOM before 0.14
so this.refs.fileInput.click() will not work and will throw TypeError in version before v0.14

### DIFF
--- a/index.js
+++ b/index.js
@@ -114,7 +114,13 @@ class Dropzone extends React.Component {
   open() {
     var fileInput = this.refs.fileInput;
     fileInput.value = null;
-    fileInput.click();
+    if (fileInput.click && typeof fileInput.click === 'function'){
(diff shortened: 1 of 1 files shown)


okonet:
v3.x is not compatible with React 0.13 on more levels (require calls, different packages etc) and I don't think it is worth trying to add compatibility for 2 versions. React 0.13 users should use v2.x instead.

ripitrust:
Thanks for your reply

I think the docs should be updated then, to avoid confusion

okonet:
You are right. I've updated `peerDependencies` in `package.json`. I'll update readme as well.

ripitrust:
great
thanks for your work

okonet:
Done!

PR #95, opened 2015-11-12
Add support for rejectStyle

indraneel:


paramaggarwal:
Thanks! Merging this in. Would be great to mention it in the docs too!

PR #97, opened 2015-11-25
add createObjectURL wrapper function

geminiyellow:
when i use phantomjs run my test, i got this error:

```
TypeError: 'undefined' is not an object (evaluating 'window.URL.createObjectURL')
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -86,7 +86,7 @@ class Dropzone extends React.Component {
 
     for (var i = 0; i < max; i++) {
       var file = droppedFiles[i];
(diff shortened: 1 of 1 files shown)


paramaggarwal:
Looks good! @okonet

geminiyellow:
@okonet ok, what about this?

okonet:
This will work in every modern browser, but will it if none of `window.URL` or `window.webkitURL` will be undefined? I think the code should also check if key `createObjectURL` exist and that it's a type `function`, right?

PR #98, opened 2015-11-26
Add disablePreview prop to support big files that don't need preview

HenriJ:
I am using this component to hash files in the browser.
Because hashing a file is a streaming process, it doesn't need to read the whole file to process it : thus I can drop bigger files in the drop zone.
Unfortunately right now the preview creation inside the component makes it impossible to handle big files (>100Mo)


okonet:
Awesome! Can you please also update the README?

HenriJ:
Sure thing

okonet:
Cool! Thanks!

PR #103, opened 2015-12-02
server side rendering fix

bruno12mota:
Fix to allow server side rendering, document var should always be checked for undefined


mull:
:+1: just ran into this

okonet:
Cool thanks!

mull:
Thanks for the swift response @okonet!

Issue #99, opened 2015-12-01
Filter by file types on dialog open


bobiblazeski:
Is it possible to filter files when Dialog is open, I would like user to see only image files: png & jpg.

okonet:
Yes, use `accept` prop to pass desired MIME types according to http://www.w3.org/TR/html-markup/input.file.html

okonet:
README should inform about it

bobiblazeski:
Works great, many thanks.

PR #105, opened 2015-12-11
Add accept documentation in README

xabikos:
Add documentation and a short example for accept option in README to solve #99


okonet:
Thanks!

PR #109, opened 2015-12-16
Support additional props on root element and file input

### LINKED ISSUE #108: Additional props
It would be helpful if any additional props that are passed to the component are rendered on the root node, rather than being ignored.

For example:
- `tabIndex`
- `aria-*`
- `hidden`
- `title`
etc, etc

Should be really easy to implement using rest/spread ops.

```
// Known props that component handles specifically are extracted, 
// everything else is added to props
let { prop1, prop2, ...props } = this.props;
...
return (
    <div prop1={prop1} prop2={prop2} {...props}>
        ...
    </div>
);
```

at0g:
Fixes #108

- Add support for dynamic props on root element
- Add support for dynamic props on input element (via inputProps prop)
- Write basic tests for dynamic props
- Added `react-testutils-additions` as a dev dependency, for $ style find selectors

- Fix babel reference in mocha-environment (could not run tests from a clean install otherwise)
- Fix babel reference in npm build script to use local installed babel

### DIFF
--- a/index.js
+++ b/index.js
@@ -124,28 +124,35 @@ class Dropzone extends React.Component {
   }
 
   render() {
-    let className, style, activeStyle, rejectStyle;
+    const {
+      accept,
+      activeClassName,
+      inputProps,
+      multiple,
+      name,
+      rejectClassName,
+      ...rest
+    } = this.props;
+    let {
+      activeStyle,
+      className,
+      rejectStyle,
+      style,
+      ...props
+    } = rest;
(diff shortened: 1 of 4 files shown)


at0g:
@okonet - Updated with your feedback, also added a couple more tests and removed the relative path to babel from npm test.
Let me know if you have any other amends.
cheers

okonet:
Looking good! Thanks!

PR #110, opened 2015-12-18
Added the option to select only Folders based on the onlyFolders option, available only for chrome 21+

SamvelRaja:
Added the option to select only Folders based on the onlyFolders option, available only for chrome 21+


SamvelRaja:
Added the option to select only Folders based on the onlyFolders option, available only for chrome 21+

PR #118, opened 2015-12-20
add unique key to image preview

SylarRuby:
We seem to need a unique key for each preview of images (but multiple not working: https://github.com/paramaggarwal/react-dropzone/issues/117)

```
// adding a range between 50 and 1 - change to your need:
var r = Math.floor((Math.random() * 50) + 1);

{this.state.files.map((file) => <img key={r} src={file.preview}
```


okonet:
I'm closing this one because of 

> SylarRuby  wants to merge 9 commits into master from gh-pages

Please create a better a new PR with this fix or create a separate issue. Also it can be simplified by

```
{this.state.files.map((file, idx) => <img key={idx} src={file.preview}...
```

afgarcia86:
This helped me resolve the warning I was getting. I would consider updating the readme

okonet:
This is outside of scope of this library and is React-specific detail which might change in the future.

PR #121, opened 2015-12-29
Support one or more children.

### LINKED ISSUE #120: Only a single child?
Is there any specific reason why only a single child is allowed? 

```
children: React.PropTypes.element
```

I've been using Dropzones with multiple children in an older version for some time and never had issues. I can of course just wrap all children in a `span` or `div`, but since the Dropzone children render in its `div` it seems redundant.

marnusw:
Fixes #120

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -237,7 +237,7 @@ Dropzone.propTypes = {
   onDragEnter: React.PropTypes.func,
   onDragLeave: React.PropTypes.func,
 
-  children: React.PropTypes.element,
+  children: React.PropTypes.node,
   style: React.PropTypes.object,
   activeStyle: React.PropTypes.object,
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -237,7 +237,10 @@ Dropzone.propTypes = {
   onDragEnter: React.PropTypes.func,
   onDragLeave: React.PropTypes.func,
 
-  children: React.PropTypes.element,
+  children: React.PropTypes.oneOfType([
```
okonet: Wouldn't it be better to just use `React.PropTypes.node`?

marnusw:
Of course, forgot about that just now. Fixed...

okonet:
Cool! Thanks!

PR #127, opened 2016-01-08
Use package babel

dduugg:
The global babel may not be compatible. (For instance, the library is not currently compatible with babel 6 because of the `.babelrc` reference to `"stage": 0`).


dduugg:
solved by upgrading npm

Issue #112, opened 2015-12-18
Use webpack or rollout to compile in order to produce more compatible version


okonet:
This will allow to build more compat ES5 version and uglify easily

okonet:
#132

PR #132, opened 2016-01-24
Update build environment (#112)

vlindhol:
Tweaked the build environment:

* Upgraded to babel6.
* Removed the "loose" parameter in babel, since it's not recommended.
Testing shows no  behaviour difference in the resulting code.
* Added a webpack step to generate a ES5 and <script> friendly module
(webpack is also a devDependency now).
* Added a uglifyjs step to minify the bundle (uglifyjs also a
devDependency now)

react-dropzone can now be hosted on a CDN and used directly in a
<script> tag, although the package attr-accept must also be referenced.
Maybe bake it into the code with webpack?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ node_modules
 
 # Generated files
 lib
+dist
 
 #IDE
 .idea
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@ logs
 node_modules
 
 # Generated files
-lib
 dist
 
 #IDE
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Simple HTML5 drag-drop zone with React.js",
   "main": "lib/index.js",
   "scripts": {
-    "build": "babel ./src --out-dir ./lib",
+    "build": "babel ./src --out-dir ./lib && webpack && uglifyjs ./dist/react-dropzone.js --compress --mangle --output ./dist/react-dropzone.min.js",
     "prepublish": "npm run lint && npm test && npm run build",
     "test": "mocha --require ./mocha-environment.js './src/test.js'",
     "lint": "eslint ./src",
@@ -42,8 +42,13 @@
     "attr-accept": "^1.0.3"
   },
   "devDependencies": {
-    "babel": "^5.8.29",
+    "babel-cli": "^6.4.5",
+    "babel-core": "^6.4.5",
     "babel-eslint": "^4.1.6",
+    "babel-plugin-add-module-exports": "^0.1.2",
+    "babel-preset-es2015": "^6.3.13",
+    "babel-preset-react": "^6.3.13",
+    "babel-preset-stage-0": "^6.3.13",
     "chai": "^3.4.1",
     "eslint": "^1.10.3",
     "eslint-config-airbnb": "^2.1.1",
@@ -55,6 +60,8 @@
     "react-addons-test-utils": "^0.14.3",
     "react-dom": "^0.14.3",
     "react-testutils-additions": "^0.2.0",
-    "sinon": "^1.17.2"
+    "sinon": "^1.17.2",
+    "uglifyjs": "^2.4.10",
+    "webpack": "^1.12.11"
   }
 }
--- a/package.json
+++ b/package.json
@@ -2,9 +2,9 @@
   "name": "react-dropzone",
   "version": "3.3.2",
   "description": "Simple HTML5 drag-drop zone with React.js",
-  "main": "lib/index.js",
+  "main": "dist/index.js",
   "scripts": {
-    "build": "babel ./src --out-dir ./lib && webpack && uglifyjs ./dist/react-dropzone.js --compress --mangle --output ./dist/react-dropzone.min.js",
+    "build": "webpack",
     "prepublish": "npm run lint && npm test && npm run build",
     "test": "mocha --require ./mocha-environment.js './src/test.js'",
     "lint": "eslint ./src",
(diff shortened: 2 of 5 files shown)


vlindhol:
I'm not very experienced with the Javascript dev stack so hopefully I did everything right :smile:

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Simple HTML5 drag-drop zone with React.js",
   "main": "lib/index.js",
   "scripts": {
-    "build": "babel ./src --out-dir ./lib",
+    "build": "babel ./src --out-dir ./lib && webpack && uglifyjs ./dist/react-dropzone.js --compress --mangle --output ./dist/react-dropzone.min.js",
```
okonet: The point of using webpack is to have the whole compiling pipeline set up with it. Otherwise it doesn't make any sense. Can you switch to webpack with uglify plugin instead of using uglifyJS directly?

webpack.config.js
```diff
@@ -1,14 +1,12 @@
...
-    externals: {
-      "react": "React"
-    }
-  }
+module.exports = {
+	entry: "./lib/index.js",
+	output: {
+		path: __dirname,
+		filename: "./dist/react-dropzone.js",
+		libraryTarget: "var",
```
okonet: I'd prefer to export UMD module since it seem to be the most compatible one.

.gitignore
```diff
@@ -7,6 +7,7 @@ node_modules
 
 # Generated files
 lib
+dist
```
okonet: Right now module being exported into `lib` so if you change it here you have to updated `package.json` as well.

vlindhol:
Made the tweaks mentioned in your feedback! Note: I also turned on source map generation, is that wanted default behaviour? The uglified file and source map are both placed in "dist" now.

webpack.config.js
```diff
@@ -1,14 +1,43 @@
...
+    module: {
+        loaders: [
+            {
+                include: [
+                    path.resolve(__dirname,"src"),
+                ],
+                test: /\.js$/,
+                loader: 'babel-loader',
+                query: {
+                    presets: ['es2015','react','stage-0'],
```
okonet: In this case we won't need a `.babelrc` file I guess.

webpack.config.js
```diff
@@ -1,14 +1,43 @@
...
-      path: __dirname,
-      filename: "./dist/react-dropzone.js",
-      libraryTarget: "var",
-      library: "Dropzone"
+var webpack = require('webpack');
+var path = require('path');
+
+module.exports = {
+	entry: "./src/index.js",
+    devtool: "source-map",
```
okonet: Is the indentation a bit off here?

webpack.config.js
```diff
@@ -1,14 +1,43 @@
...
+        // Can require('file') instead of require('file.js') etc.
+        extensions: ['','.js','.json']
+    },
+	externals: {
+		"react": "React"
+	},
+    plugins: [
+        new webpack.optimize.UglifyJsPlugin({
+            compress: {
+                // does this actually do anything?
```
okonet: It prevent UglifyJS from outputting a lot of warnings.

Also, we might want `screw_ie8` option be set to true here (check the uglifyJS options to make sure the name is correct)

webpack.config.js
vlindhol: Actually the testing framework still uses it, I couldn't figure out how to give the babel parameters directly to mocha.

webpack.config.js
```diff
@@ -1,14 +1,43 @@
...
-      library: "Dropzone"
+var webpack = require('webpack');
+var path = require('path');
+
+module.exports = {
+	entry: "./src/index.js",
+    devtool: "source-map",
+	output: {
+		path: __dirname,
+		filename: "./dist/react-dropzone.js",
```
okonet: I'd prefer `index.js`?

webpack.config.js
```diff
@@ -1,14 +1,43 @@
...
-      libraryTarget: "var",
-      library: "Dropzone"
+var webpack = require('webpack');
+var path = require('path');
+
+module.exports = {
+	entry: "./src/index.js",
+    devtool: "source-map",
+	output: {
+		path: __dirname,
```
okonet: Why not setting this to `./dist` and removing it from filename?

okonet:
Looks awesome! I think the `react-dom` is still missing in the `externals`. Did you check the build size?

webpack.config.js
okonet: In this case, let's just use `.babelrc` and remove `query` here.

vlindhol: In other words it might be better to go back to just using `.babelrc`, I thought webpack needed the parameters specified in `webpack.config.js` but that does not seem to be the case. I'll update the PR.

okonet: Yes, when you use same babel config in different places, it's better to use `beblrc` or `package.json` files. If you want override it, this is how you do it. See this for example: https://github.com/mxstbr/react-boilerplate/blob/v3.0.0/webpack/webpack.dev.babel.js#L44-L47

vlindhol:
Adding react-dom to the externals didn't change the build size. As far as I can see it is not referenced anywhere else so webpack doesn't include it even if it is a dev dependency. Can that dependency be removed, or should it be made an official dependency?

okonet:
Ok cool. I'll check out the deps.

okonet:
This one should close #112

Lucretius:
@okonet Any idea as to when this PR is going to get merged?  The outdated babel configuration causes ReferenceErrors (Using removed Babel 5 option) for newer webpack projects.

okonet:
@Lucretius I don't have much time now, so I'm open for the PR.

PR #139, opened 2016-02-15
Fixing an issue when the droppedFiles array is empty

cheton:
I'm using both [Sortable](https://github.com/RubaXa/Sortable) and [react-dropzone](https://github.com/okonet/react-dropzone) in the web interface, and made the whole body a dropzone like below:

However, when I drag and drop sortable widgets only within the UI, but not dragging files from somewhere else, it will still invoke react-dropzone's `onDrop` method, and generated an error:
```
Uncaught TypeError: Cannot set property 'preview' of undefined
```

Since there are no files, it should stop proceeding and return immediately if the `droppedFiles` array is empty.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -84,7 +84,7 @@ class Dropzone extends React.Component {
     const droppedFiles = e.dataTransfer ? e.dataTransfer.files : e.target.files;
     const files = [];
 
-    if (droppedFiles.length === 0) {
+    if (!droppedFiles || (droppedFiles.length === 0)) {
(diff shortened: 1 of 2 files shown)


okonet:
That's awesome! Thanks! 

Could you please also add tests for it?

cheton:
I just added a simple test. You can do the verification again. Thanks.

okonet:
I'll close this one in favor of #156. Thanks for the contribution!

PR #142, opened 2016-02-28
Add more MIME type examples

### LINKED ISSUE #141: Accept multiple MIME types?
I would like my instance of Dropzone to only accept images, is there a way to do that? Since `accept` should be a string, therefore a single MIME type is supported.

silvenon:
They can be a single MIME type or a comma delimited list. It's not something that I could easily find without some digging, so I thought it would be useful to provide this shortcut info right away.

Closes #141.


okonet:
Thanks!

PR #150, opened 2016-03-10
Add `global` prop

olalonde:
When the global property is set to true, the whole browser window becomes the drop target. It attaches event listener on document when it mounts and removes them when it unmounts.

Here's a usage example:

``` jsx
import React, { Component } from 'react'
import DropZone from 'react-dropzone'
import classnames from 'classnames'

export default class DragDrop extends Component {
  constructor(props) {
    super(props)
    this.onDrop = this.onDrop.bind(this)
    this.onDragEnter = this.onDragEnter.bind(this)
    this.onDragLeave = this.onDragLeave.bind(this)
    this.state = {
      isDragActive: false,
    }
  }

  onDragEnter() {
    if (!this.state.isDragActive) {
      this.setState({ isDragActive: true })
    }
  }

  onDragLeave() {
    if (this.state.isDragActive) {
      this.setState({ isDragActive: false })
    }
  }

  onDrop(files) {
    this.setState({
      isDragActive: false,
    })
    console.log('do something with files')
  }

  render() {
    const { children } = this.props
    const { isDragActive } = this.state

    return (
      <div>
        <DropZone
          multiple={false}
          global
          disableClick
          style={{ display: 'none' }}
          activeStyle={{ display: 'none' }}
          onDrop={this.onDrop}
          onDragEnter={this.onDragEnter}
        />
        <div className={classnames({ hide: !isDragActive })}>
          <div className="modal-backdrop fade in"></div>
          <div className="fade in modal" style={{ display: 'block' }}>
            <div className="dragdrop">
              <div className="message">
                <h1 className="title">Drop to file to upload...</h1>
              </div>
            </div>
          </div>
        </div>
        <div>
          {children}
        </div>
      </div>
    )
  }
}
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -19,8 +19,26 @@ class Dropzone extends React.Component {
     };
   }
 
+  componentWillMount() {
+    if (this.props.global && typeof document) {
+      document.addEventListener('dragenter', this.onDragEnter);
+      document.addEventListener('dragover', this.onDragOver);
+      document.addEventListener('dragleave', this.onDragLeave);
+      document.addEventListener('drop', this.onDrop);
+    }
+  }
+
   componentDidMount() {
     this.enterCounter = 0;
+  }
+
+  componentWillUnmount() {
+    if (this.props.global && typeof document) {
+      document.removeEventListener('dragenter', this.onDragEnter);
+      document.removeEventListener('dragover', this.onDragOver);
+      document.removeEventListener('dragleave', this.onDragLeave);
+      document.removeEventListener('drop', this.onDrop);
+    }
   }
 
   onDragEnter(e) {
@@ -203,6 +221,8 @@ class Dropzone extends React.Component {
       inputAttributes.name = name;
     }
 
+    if (this.props.global) return null;
+
     return (
       <div
         className={className}
@@ -227,7 +247,8 @@ class Dropzone extends React.Component {
 Dropzone.defaultProps = {
   disablePreview: false,
   disableClick: false,
-  multiple: true
+  multiple: true,
+  global: false
 };
 
 Dropzone.propTypes = {
@@ -251,7 +272,9 @@ Dropzone.propTypes = {
   inputProps: React.PropTypes.object,
   multiple: React.PropTypes.bool,
   accept: React.PropTypes.string,
-  name: React.PropTypes.string
+  name: React.PropTypes.string,
+
+  global: React.PropTypes.bool
 };
 
 export default Dropzone;
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ npm install --save react-dropzone@2.x
 Usage
 =====
 
-Simply `require('react-dropzone')` and specify an `onDrop` method that accepts an array of dropped files. 
+Simply `require('react-dropzone')` and specify an `onDrop` method that accepts an array of dropped files.
 
 By default, the component picks up some default styling to get you started. You can customize `<Dropzone>` by specifying a `style` and `activeStyle` which is applied when a file is dragged over the zone. You can also specify `className` and `activeClassName` if you would rather style using CSS.
 
@@ -69,6 +69,7 @@ Features
 
 - `disableClick` - Clicking the `<Dropzone>` brings up the browser file picker. To disable, set to `true`.
 - `multiple` - To accept only a single file, set this to `false`.
+- `global` - To make the whole browser window become the drop target, set this to `true`.
 - `accept` - Filters the file types that are valid. It should have a valid MIME type according to [input element](http://www.w3.org/TR/html-markup/input.file.html), for example:
   * `application/pdf`
   * `image/*`


src/index.js
```diff
@@ -19,10 +19,28 @@ class Dropzone extends React.Component {
     };
   }
 
+  componentWillMount() {
+    if (this.props.global && typeof document) {
+      document.addEventListener('dragenter', this.onDragEnter);
```
okonet: You assign event listeners to some class methods that doesn't exist. This will not work I think.

okonet:
Hey! Thanks for this PR. I think this should not be a prop on the `dropzone` but a decorator/HOC. In this case it would be easier to reuse the `dropzone` and add this new "global" behavior to it. What do you think?

Also, we will need tests for this as well.

src/index.js
olalonde: What class methods that don't exist? You mean `document`? This code works in Chrome but there might be a more better cross browser solution?

olalonde:
I expected this PR would meet resistance :) With the approach you described, how would the parent component fire the drag events from Dropzone? I could indeed implement this as a completely different component but there'd be a lot of shared logic with this one (except maybe for the render method).

src/index.js
okonet: Nevermind, the top of the file was collapsed :)

okonet:
That's a good point. But I still don't like how much boilerplate your approach requires. What if we split the original `Dropzone` into 2 components: one (smart) that handles events and another (dumb) for rendering? In this case we could swap the dumb components dependeing on `global` prop or even allow custom renderers. ATM I don't like the API with `inputProps` and stuff at all. This will require more work and I'd add you as a contributor so you can commit directly to this repo. What do you think?

olalonde:
Oh got it, that would indeed make things cleaner. In fact the render part in my opinion should be left entirely to the user with maybe a default styled component in case the component has no child. I'm not sure I'll have time to work on this but could maybe give it a stab this weekend.

mikecousins:
I'm pretty interested in this as I have a need for this same functionality. Any idea what way you guys are leaning and what the timeline is yet?

aight8:
Is this project death?

okonet:
@aight8 I don't have time to work on it right now. So I'm happy to accept PRs. So if you feel you would like to work on the discussed solution — you're welcome.

okonet:
I'm still working on the documentation for this but this can be achieved without introducing new props. See http://react-dropzone.netlify.com/#FullScreen

oyeanuj:
@okonet I tried to use the method suggested in examples but ran into a major issue. It works only when you enter the dropzone (which can be a small part of the page), since only then can you drop it anywhere. In comparison, the solution offered in this PR might be a cleaner way of making the document as the drop-target. Thoughts?

@olalonde What did you end up doing finally to achieve the functionality?

olalonde:
@oyeanuj it's been a while but I think I just used the PR I had originally submitted https://github.com/okonet/react-dropzone/pull/150/files

okonet:
I'm not sure what the issue is. Why can't you wrap the whole app into the 🎯 react-dropzone if you need this functionality?

oyeanuj:
@okonet Right, that was my first impulse as well but it gets messy quickly. Here are the things that were not clean about that -

1. Firstly, the module which actually requires dropzone might be nested fairly down in the hierarchy, so you have to communicate with the App-level dropzone somehow (redux or context, or something).

2. Then, often times you might have other components using dropzone which might not require this functionality or might have other differing props being passed to dropzone. So, either you have that dropzone under the app-level dropzone, or you again communicate somehow to the app-level dropzone.

3. This is also true when you just need a click-to-open button (for which I am using dropzone) - we again then run into a dropzone inside the app-level dropzone.

4. Then, there are times when some other component might be using `onDrop` handler, and one has to ensure that event is not propagating further.

Essentially, it wasn't clean and fraught with issues. 

I totally understand if you think this should be a userland feature (although folks would be duplicating the logic), but fwiw, I think the solution proposed here seemed the cleanest.

okonet:
I might be missing something so it would be helpful to see some real-life example. Can you share it here or make OSS?

mrichmond:
I've been quietly watching on the sidelines for several months on this issue, but I'll jump in here. I completely agree with @okonet and I believe I have a similar use case. A global flag would be a huge feature for me. @oyeanuj, I'll do my best to generally describe my use case and please ask for any clarification you'd like. For what it's worth... of course I can make it work as is... but it's very dirty and I really don't like how shapes the dom tree. For my application I don't really much need a visual representation of the drag area in most cases. I have a nav bar at the top of my application with an upload button that triggers the dropzone upload. I have multiple routes in my application... something like:

`/route1 // enable nav with upload button & enable window drag and drop (no ui)`
`/route2 // enable nav with upload button & DISABLE window drag and drop (no ui)`
`/route3 // DISABLE nav with upload button & enable window drag and drop (no ui)`
`/route4 // DISABLE nav with upload button & DISABLE window drag and drop (no ui)`

(new use case route I'm adding right now to my application:)
`/route5 // DISABLE nav with upload button & enable window drag and drop & display drag and drop ui feedback`

Right now my app looks something like this:

```
<Router>
  <App>
    <Upload>
      <Dropzone>
       <div id="dropzone">
        <main>
          <Nav>
            <nav>
            <UploadButton>
             <button />
            </UploadButton>
            </nav>
          </Nav>
          <Page1>
            <section>
              <p>Hello World!</p>
            </section>
          </Nav>
        </main>
        <input type="file" />
       </dv>
      </Dropzone>
    </Upload>
  </App>
</Router>
```

This mean in my App.js I'm wrapping my nav and page in an upload component which I don't really like. Now my Upload component is loosly coupled to my Upload component and if I am on a route like "route4" where I don't need uploading then if I need some really weird logic in my App to conditionally wrap my nav and page and also my dom tree will change. I also don't like that my <main> is wrapped in a <div> ... ew. I would greatly prefer this:

```
<Router>
  <App>
    <main>
      <Upload>
        <Dropzone global>
          <div id="dropzone">
            <input type="file" />
          </div>
        </Dropzone>
      </Upload>
      <Nav>
        <nav>
          <UploadButton>
           <button />
            </UploadButton>
            </nav>
          </Nav>
          <Page1>
            <section>
              <p>Hello World!</p>
            </section>
          </Nav>
        </main>
       </dv>
      </Dropzone>
    </Upload>
  </App>
</Router>
```

oyeanuj:
@okonet Sorry I missed your above message. A real-life example is Slack - if you drag a file onto any part of Slack (including the left/right side-bar), Slack detects a drag is happening and you can drop it anywhere in the window - it automatically binds it to conversation that you are in. This is where `global` prop is helpful. A contrasting example is say Google Drive, where the drag or drop has to be in a specific area/component.

oyeanuj:
@okonet Sorry I missed your above message. A real-life example is Slack - if you drag a file onto any part of Slack (including the left/right side-bar), Slack detects a drag is happening and you can drop it anywhere in the window - it automatically binds it to conversation that you are in. This is where `global` prop is helpful. A contrasting example is say Google Drive, where the drag or drop has to be in a specific area/component.

@mrichmond Sorry, I am not sure I am following correctly - are you suggesting that the `global` prop is not needed in your case? If so, I agree. I think the `global` prop is helpful when you want the whole window to listen to drag or drop events - and want that to be declared at the component level, rather than the App level.

mrichmond:
@okonet thanks for considering this and if you want to chat more about it please let me know. I'm considering cloning the PR from @oyeanuj for now...

TL;DR ... Wrapping my entire application in a dropzone div is pretty lame ...
```
<body>
  <div id="dropzone">
    <main>
        <nav>...</nav>
        <section>...</nav>
    </main>
    <input type="file" />
  </div>
</body>
```

mrichmond:
@oyeanuj thanks for the quick reply... maybe you misunderstood. I'm saying that while technically it "works" it's ugly and I'm itching to refactor it because it feels like the currently implementation is not at all optimal. I really don't like wrapping my application in the dropbox component. It feels completely backwards to me. my Upload component (using react-dropbox) shouldn't wrap my nav and page. I just don't want to push my components through the dropbox component. If I decided that I don't need uploading on a particular route then I need some unusual logic in my app that inherently changes my dom tree between pages. That just feels so wrong to me.

src/index.js
```diff
@@ -203,6 +221,8 @@ class Dropzone extends React.Component {
       inputAttributes.name = name;
     }
 
+    if (this.props.global) return null;
```
mrichmond: I'd have to pull this down and test it... but I don't think you want to necessarily return null here. If global (or whatever the property is called) is specified I'd still like to render the dropzone div with the input. If preventing the output is a requirement for you, I'd suggest separating that into another property. I can certainly see a use case for having a drop area with content that changes whenever a file is dragged over the entire window.

oyeanuj:
@mrichmond Got it, thanks for clarifying. I think we are both in the same boat, with same concerns then.

mrichmond:
Yup... and @olalonde got the ball rolling with his PR. I think just a few tweaks and it's there, but that's just my $0.2 on the matter.

okonet:
Re wrapping the whole app into the 🎯 React-dropzone: I don't see how this is ugly if the whole app of yours can accept files? Slack example is a perfect one and this is totally possible to do now with the current implementation.

My concerns about the `global` props:

1. The behaviour is very implicit and hidden in the tree somewhere in the app although the root level DOM element is the one who will receive events.
1. What is gonna happen if you have multiple dropzone components with `global` prop set to `true` on a single page? There are no guarantees that this isn't gonna happen and I'm not sure how it will behave in this case. Lots of potential bugs are here IMO.
1. You don't have to modify the DOM tree nor remove dropzone from the React tree if it should be disabled on some routes. You can just do no do anything in the `onDrop` based on your route. 
1. I'll gladly accept a PR that will introduce `disabled` prop to the 🎯 react-dropzone that will disable all event handling and thus UI updates for it.

If I'm still miss some edge cases, I'm all ears but I don't think you're convincing me to reopen this PR yet.

oyeanuj:
@okonet I think the distinction that you are missing is not that the 

>  if the whole app of yours can accept files

but that the whole window can accept files when on a certain screen. In other words, when you are lets say creating a new channel on Slack, dragging or dropping files doesn't do anything. But when you are in a channel conversation, then it activates. 

In this scenario, with current Dropzone's functionality, it would mean that you are toggling an app level property from fairly deep within the app - which feels weird.

--- 

To your questions, some thoughts - 
> 2. What is gonna happen if you have multiple dropzone components with global prop set to true on a single page? There are no guarantees that this isn't gonna happen and I'm not sure how it will behave in this case. Lots of potential bugs are here IMO.

Do you think this can be solved with the idea that the last one would override the others - similar to how one expects with variable assignment. In most cases, this is probably going to be by accident, and for cases where this is by design, I assume developers can build solution on top of it - like they would if they weren't using `Dropzone`.

> 2. You don't have to modify the DOM tree nor remove dropzone from the React tree if it should be disabled on some routes. You can just do no do anything in the onDrop based on your route.

I think the bigger problem is that the logic for enabling/disabling can become complicated and forces your root level to know a lot of internal component logic or detail or use Redux state to globally enable/disable - which then becomes similar to just having a `global` prop.

okonet:
> toggling an app level property from fairly deep within the app - which feels weird.

Well, that's why Facebook came up with flux architecture. This is very common in the React world.

> the last one would override the others

I'm not sure what you mean by that. Dropzone doesn't hold any state and will not know about "others" instances. This have to happen in the user land and this is fairly complex.

> which then becomes similar to just having a global prop.

I think you're trying to push some of user's (i.e. developer's) responsibilities to this project which I think is a bad design-wise. Sorry, but till now this discussion wasn't convincing for me. If you feel strongly about it, you can make a fork and implement it the way you want but I believe I won't accept this to be part of the main repo.

okonet:
@oyeanuj I've just merged https://github.com/react-dropzone/react-dropzone/pull/473 that implements `disabled` prop

oyeanuj:
@okonet thanks, will check it out. Even though we have disagreement on the solution, appreciate you having the discussion!

kngroo:
Sorry to revive this, but I'm running into an issue that seems to be related to the topic of this thread.
I am trying to use dropzone to wrap the root component that sets out the layout of my whole application.

ex:
```
<RootComponent>
  <DropZone>
    <Header />
    {props.children}
    <Footer/>
  </DropZone>
</RootComponent>
```

If I wanted to include a button in my child component that has a `<button onClick={e => dropzoneRef.open() } />`, there is no way to obtain the `dropzoneRef` without calling something like `findDOMNode` or passing the prop down multiple layers, or keeping it in a redux store.

I attempted to fix it by removing `<DropZone>` from the root component and putting it inside my child components, and applying appropriate styles to the `.file-drop` class to cover the entire window. 

ex:
```
<ChildComponent>
  <DropZone 
    style={{position: 'fixed', top: 0, left: 0, width: '100%', height: '100%', zIndex: 1}}
    ref={el => dropzoneRef = el} />
  {components}
</ChildComponent>
```

The problem with the above approach is that if you want Dropzone to detect the drag event, you need to set the z-index so that the Dropzone container div is in front of all my other child components. This prevents interaction with the other components in the children.

If there were something like a global prop that binds the drag event listener to the document, the above approach would work.

Thoughts?

PR #156, opened 2016-03-30
Fix off-by-one in droppedFiles length

### LINKED ISSUE #155: Dropped text/uri-list or text/html causes exception if the multiple prop is false
If a dropzone is set to `multiple={false}` and a user drops something that is not an image file onto it (for example, an image from within their browser window, or a link), it causes a javascript error. This occurs even if the item doesn't match the `accept` property and so should be rejected. Here is an example: http://www.webpackbin.com/41KCncGRe

This doesn't occur if disablePreview is true.

In Firefox:

`TypeError: Argument 1 is not valid for any of the 2-argument overloads of URL.createObjectURL.`

In Chrome:

`Uncaught TypeError: Failed to execute 'createObjectURL' on 'URL': No function was found that matched the signature provided.`

BJTerry:
Fixes #155

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -82,7 +82,7 @@ class Dropzone extends React.Component {
     });
 
     const droppedFiles = e.dataTransfer ? e.dataTransfer.files : e.target.files;
-    const max = this.props.multiple ? droppedFiles.length : 1;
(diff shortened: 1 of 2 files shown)


okonet:
Awesome! Thanks a lot!

PR #160, opened 2016-04-04
fix #159 require 'React' instead of 'react'

### LINKED ISSUE #159: Should require 'react' (lower case)
I'm getting warnings from Webpack when building my project related to dropzone requiring 'React' instead of 'react'.

`There is another module with an equal name when case is ignored.`

j-san:
fix issue `Cannot find module "React"`

https://travis-ci.org/okonet/react-dropzone/builds/120584465#L205

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -26,7 +26,7 @@ module.exports = {
     extensions: ['', '.js', '.json']
   },
   externals: {
-    react: 'React'
+    react: 'react'
   },
   plugins: []
 };


okonet:
Thanks! :+1:

Mistereo:
It seems to me that this will not work for browsers (without commonjs/requirejs environments).

okonet:
It never was and never will be I think. It just not common for react libs to have React bundled into the dist build. Please use webpack/babelify.

j-san:
You're welcome :smile:

PR #163, opened 2016-04-07
Manual upload handler clarity

phoenixbox:
The `onDrop` function is used as the handler for both the `drop` and `click` upload actions. The handler name slightly conflicts across actions, so I wasn't sure if there was an undocumented handler for the manual `open` action. When I looked at the source I could see that `onDrop` was used for both.

As a developer I think I would have found it useful to see that the `onDrop` handler was also the completion handler for the manual open action so I added it to the README.

Thanks for the lib its great! ❤️ ❤️ ❤️

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ npm install --save react-dropzone@2.x
 Usage
 =====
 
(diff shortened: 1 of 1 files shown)


phoenixbox:
I will make a PR on the `gh-pages` branch also :)

PR #173, opened 2016-05-11
If accept prop is specified, enforce it on drop

### LINKED ISSUE #138: onDrop is being called with files that shouldn't be accepted
IMO, if the file isn't accepted, it shouldn't be sent to the onDrop method

dnd

### LINKED ISSUE #145: Accpets files dropped of the wrong MIME type?
The CSS changes to red but if a different file is dropped on it it still accepts it?

tmarshall:
closes #138
closes #145

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-dropzone",
-  "version": "3.4.0",
+  "version": "4.0.0",
   "description": "Simple HTML5 drag-drop zone with React.js",
   "main": "dist/index.js",
   "scripts": {
--- a/src/index.js
+++ b/src/index.js
@@ -94,11 +94,11 @@ class Dropzone extends React.Component {
       files.push(file);
     }
 
-    if (this.props.onDrop) {
-      this.props.onDrop.call(this, files, e);
-    }
-
     if (this.allFilesAccepted(files)) {
+      if (this.props.onDrop) {
+        this.props.onDrop.call(this, files, e);
+      }
+
       if (this.props.onDropAccepted) {
         this.props.onDropAccepted.call(this, files, e);
       }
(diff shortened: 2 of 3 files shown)


okonet:
Hey!thanks for the PR! Looking good. 

I noticed you changed the logic a bit that isn't covered with tests. Do you mind adding tests for cases when `accepts` isn't specified? If they will be green, I'm good to merge this. 

I'm also wondering if this one will require a major version bump because of this change. What do you think?

tmarshall:
> I'm also wondering if this one will require a major version bump because of this change. What do you think?

Seems like a patch to me, since it is resolving an interaction, as to how the developers expect it to happen. Though it depends on your perspective

tmarshall:
@okonet test added

okonet:
Yes but isn't it a breaking change? I mean for those who are relying on a wrong behavior it will break stuff.

tmarshall:
@okonet from that perspective yes :)

alexeygolev:
Any progress on merging this?

okonet:
I'll release a patch version

hank7444:
ok, I found the issue, `onDrop is now called only if all files are passing the accept`,
I commented my accept and it worked

hank7444:
my accept is java .jar file, it not works :(

accept="application/java-archive"

okonet:
I'm not sure `application/java-archive` is a valid parameter. See the spec: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#Attributes

hank7444:
it seems valid 
[https://en.wikipedia.org/wiki/JAR_(file_format)](https://en.wikipedia.org/wiki/JAR_(file_format))

hank7444:
## mac chrome
image

## window 7 chrome, it type is empty
image

okonet:
Okay, so it's a browser issue after all. Use an extension pattern as a work around.

hank7444:
OK, I will try it, thx a lot!

PR #177, opened 2016-05-22
Bind this.onDrop to use desired component

seppo0010:


okonet:
Hey! Thanks for you contribution! You don't need to use `bind` when using `createClass`. You could update the example to use `class`, but even in this case, this would be an anti-pattern. It's much better performance-wise to bind before and use the reference to the bound function in the `render` method to pass it down.

seppo0010:
Uhm, I'm sorry then. I tried this example (with a modification since I was using babel (like in this [example](https://babeljs.io/blog/2015/06/07/react-on-es6-plus)) and it didn't work without the `bind`. What do you recommend me in this case? should I use `bind` or is there another alternative? Thanks

okonet:
You could bind the method in the constructor:

```jsx
class MyComponent extends React.Class {
    constructor() {
        this.onDropCallback = this.onDrop.bind(this);
    }

    onDrop() {
    }
}
```

seppo0010:
Thanks.

PR #187, opened 2016-06-16
only publish the dist folder on npm

mathieudutour:


okonet:
Why do you think this is better than publishing source code and readme? What's the issue now?

mathieudutour:
It is publishing the readme anyway. But why do you want to make everybody download the source, the travis file, etc.?

mathieudutour:
Also I'm having a weird bug because the .babelrc is included

okonet:
I often look into source code of other libraries, esp when working offline.

Not sure what issues you're having but not having `.babelrc` in just this one package won't solve the issue for you I guess.

mathieudutour:
it actually does

mathieudutour:
this is the only library which is publishing the `.babelrc` (over 30ish libraries i'm using)

okonet:
That is just a coincidence in your case and it won't solve your issue. 

That's all `.babelrc` files inside this package's `node_modules`:

```bash
$ find node_modules -name '.babelrc' 
node_modules/attr-accept/.babelrc
node_modules/babylon/.babelrc
node_modules/depcheck/.babelrc
node_modules/escope/.babelrc
node_modules/eslint-plugin-jsx-a11y/.babelrc
node_modules/estraverse/.babelrc
node_modules/jsx-ast-utils/.babelrc
```

mathieudutour:
It does solve my issue: 
`Unknown plugin "add-module-exports" specified in ".../node_modules/react-dropzone/.babelrc" at 0, attempted to resolve relative to ".../node_modules/react-dropzone" while parsing file: .../node_modules/react-dropzone/dist/index.js`
. Because it's the only one requiring the `add-module-exports` plugin.

Anyway, I'll just continue to rm the .babelrc file on post install and download a twice as big module...

okonet:
Why are you transpiling installed packaged in the first place?

mathieudutour:
I'm not transpiling, that's the weird thing. But I guess it's more a browserify/babelify issue than react-dropzone.

> Regarding the package size: I don't think that transferring another 20KB is an issue, is it? By doing so I allow developers to look up the source code and tests for the package, which is often acts better as any documentation.

That's one way to look at it. The other way is that we end up with gigantic node_modules directories which could be a lot smaller. 
But it's not a big deal for me, I have a good connection.

okonet:
I'm not familiar with browserify so I can't help but it is definitively an issue with it. In webpack, you just exclude `node_modules` from babel loader.

tleunen:
If you wanted to let developers see the source files as well, you could easily just push `src` and `dist` then ;)

okonet:
And also manually add README and license and changelog. And babelrc and other required stuff in case someone wants to build from source.

On a serious note: I don't get the issue with publishing the source code of the package on npm and won't perform any actions until I'm convinced this is bad.

Issue #193, opened 2016-07-03
Unknown props in React v15.2


Khrall:
Just noticed this warning when using version 15.2 of React:

```
Warning: 
Unknown props `disablePreview`, `disableClick` on <div> tag. 
Remove these props from the element. 
For details, see https://fb.me/react-unknown-prop
    in div (created by Dropzone)
    in Dropzone (created by xxx)
```

The [link provided](https://fb.me/react-unknown-prop) seems to indicate that the problem is at line 219 of index.js.

```
216 <div
    ...
219 {...props /* expand user provided props first so event handlers are never overridden */}
    ...
226 >
```

Anyway, I don't know much about this stuff, might be I screwed up something else. Just wanted to let you know in case it's going to show up for other people (:

OlivierCo:
<img width="1398" alt="screen shot 2016-07-04 at 12 32 58" src="https://cloud.githubusercontent.com/assets/1804322/16559234/78594f68-41e3-11e6-80c9-ffa1b1af64c4.png">

okonet:
Fixed in #194

okonet:
Please update to [v3.5.2](https://github.com/okonet/react-dropzone/releases/tag/3.5.2)

mikecousins:
Beautiful, thanks for the quick work.

okonet:
A follow up fix in #197

PR #194, opened 2016-07-04
Fix unknown props warning in React v15.2

nuc:
Relevant issue: https://github.com/okonet/react-dropzone/issues/193

Not sure if that's the _traditional_ way to go. 

I also fixed a couple additional eslint warnings in the meantime.

Feedback is welcome :)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -216,7 +216,7 @@ class Dropzone extends React.Component {
       <div
         className={className}
         style={appliedStyle}
-        {...props /* expand user provided props first so event handlers are never overridden */}
+        {...props/* expand user provided props first so event handlers are never overridden */}
         onClick={this.onClick}
         onDragStart={this.onDragStart}
         onDragEnter={this.onDragEnter}
@@ -226,7 +226,7 @@ class Dropzone extends React.Component {
       >
         {this.props.children}
         <input
-          {...inputProps /* expand user provided inputProps first so inputAttributes override them */}
+          {...inputProps/* expand user provided inputProps first so inputAttributes override them */}
           {...inputAttributes}
         />
       </div>
--- a/src/index.js
+++ b/src/index.js
@@ -149,8 +149,6 @@ class Dropzone extends React.Component {
       className,
       rejectStyle,
       style,
-      disablePreview, // eslint-disable-line no-unused-vars, prefer-const
-      disableClick, // eslint-disable-line no-unused-vars, prefer-const
       ...props // eslint-disable-line prefer-const
     } = rest;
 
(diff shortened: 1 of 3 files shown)


okonet:
Looks like this breaks the tests...

nuc:
@okonet It seems that in the new react version, the `name` attribute is always present:

https://esnextb.in/?gist=b0122d98b384e9143eadfd558228be45

So I'm not sure if [that failing test](https://github.com/okonet/react-dropzone/blob/master/src/test.js#L178-L183) makes sense anymore.

okonet:
@nuc thanks for researching it. I've reported an issue to https://github.com/facebook/react/issues/7198. Let's see what they think about it. In the meantime, I'll skip the test and merge your PR.

Thanks again for contributing!

src/index.js
```diff
@@ -149,6 +149,8 @@ class Dropzone extends React.Component {
       className,
       rejectStyle,
       style,
+      disablePreview, // eslint-disable-line no-unused-vars, prefer-const
+      disableClick, // eslint-disable-line no-unused-vars, prefer-const
```
okonet: Since these variables aren't used here I'm wondering if it is better to use the last method with `delete` props from the cloned object here:

```
function MyDiv(props) {

  const divProps = Object.assign({}, props);
  delete divProps.layout;

  if (props.layout === 'horizontal') {
    return <div {...divProps} style={getHorizontalStyle()} />
  } else {
    return <div {...divProps} style={getVerticalStyle()} />
  }
}
```

See https://facebook.github.io/react/warnings/unknown-prop.html

What do you think? Mind updating the PR? Also, could you add a test that checks that these props aren't removed from the original `this.props`?

nuc:
Sure, will do!

okonet:
I've submitted a fix to the `name` regression to react: https://github.com/facebook/react/pull/7199

nuc:
There was no need to add test for `disablePreview`, since that case was already covered in the current test: `disablePreview` is used inside `onDrop` which happens after we delete the property in `render`.

Added two small tests for `disableClick` though.

src/index.js
```diff
@@ -214,11 +212,15 @@ class Dropzone extends React.Component {
       inputAttributes.name = name;
     }
 
+    const divProps = Object.assign({}, props);
```
okonet: Let's re-write this as `const divProps = { ...props };` for consistency.

okonet:
> Added two small tests for disableClick though.

Awesome!

nuc:
Should be ready now! :)

okonet:
v3.5.2 is 🚀

elado:
Thanks for the PR, @nuc. I'm still seeing `Unknown prop `onDropAccepted` on <div> tag.`

okonet:
Looks like not all props are removed...

nuc:
Interesting, I encountered only the disablePreview and disableClick.

I can prepare a small patch tomorrow if you want.

Cheers

okonet:
@nuc that would be great!

nuc:
@okonet Added additional PR. Cheers!

Issue #104, opened 2015-12-10
Validating files and disabling preview


jeron-diovis:
How it's supposed to validate, for example, file size? Or separate valid files from invalid ones?

How about introducing smth like `validate` callback? We can pass each dropped file there, split them by returned result and then pass corresponding arrays to `onDropAccepted` and `onDropRejected`.

---

Also, it's said "We might want to disable the preview creation to support big files", but there is no way to detect whether file is actually "big". So we should either disable previews at all, or create previews for everything. Maybe, it's worth to add another function to check whether preview for particular file should be created?

---

And, shouldn't it be `this.props.disablePreview` [here](https://github.com/paramaggarwal/react-dropzone/blob/master/index.js#L93)?

okonet:
Do you mind creating a pull request?

jeron-diovis:
For now I'm just asking about these changes. If you have no objections about them, I will submit a PR.

okonet:
Yeah, it totally makes sense to me. Please submit your PR and we can discuss it from there. Thanks!

YannPl:
is this issue abandoned, I would like to have this option too =)

h8rry:
+1 on validating size

thecaddy:
+1

stueynet:
This would be pretty handy. Really would help to match this with the server limitations otherwise stuff just fails.

MichaelLeeHobbs:
Here's how I validated for anyone finding this.

      const Preview = (props) => (
        (props.file) ?
          <div>
            <img style={{width: '196px', height: '196px'}} src={props.file.preview}/>
          </div> : <div>{props.text}</div>
      );

      const TWO_MiB = 2097152;
      export class ImageDropZone extends React.Component {
        constructor(props) {
          super(props);
          this.maxSize = props.maxSize ? props.maxSize : TWO_MiB;
          this.maxWidth = props.maxWidth ? props.maxWidth : 3000;
          this.maxHeight = props.maxHeight ? props.maxHeight : 3000;
          this.text = props.text ? props.text : `Drop image here. Max ${this.maxSize} and ${this.maxWidth}px by ${this.maxHeight}px`;
      
          this.state = {
            file: null
          }
        }
      
        onDrop(files) {
          if (files.length === 0) {
            Bert.alert('No files found!', 'danger');
            return;
          }
          let file = files[0];
      
          if (file.size > this.maxSize) {
[...]

PR #202, opened 2016-07-25
added maxSize and minSize validation #104

lemuelbarango:
fix for file validation #104

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -103,7 +103,7 @@ class Dropzone extends React.Component {
       files.push(file);
     }
 
-    if (this.allFilesAccepted(files)) {
+    if (this.allFilesAccepted(files) && this.allFilesMatchSize(files)) {
       if (this.props.onDrop) {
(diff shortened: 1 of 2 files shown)


okonet:
That's awesome! Thanks for contributing. To be honest, I think this not be a part of the library but should live in a userland and this lib should allow it by providing a callback. Some users might want size validation but the other part might want to do something more complex or different. This doesn't scale ATM.

My idea was to have a callback function that users can pass along with other props and it should get called in `onDrop` function. I might get some time to work on it soon-ish, but if you feel like it, please go for it!

Does it make sense?

prabhu:
Great PR. I disagree. This should be part of the library. Callback for customising max and min size is frankly an overkill.

okonet:
@lemuelbarango can you please also update the README?

lemuelbarango:
Yeah I can

albertkawmi:
@lemuelbarango @okonet thanks for this. Will this update be published on npm?

PR #203, opened 2016-07-27
Add missing bind(this) on button tag in README

Sonblind:
Add `onClick={this.onOpenClick.bind(this)` to get the example in the README to work.


okonet:
Examples are using `createClass` that has auto-binding so this isn't needed. 

Also for perf reasons this is an anti-pattern. You should add binding in the constructor if you use es2015 classes.

Sonblind:
Ah, that makes sense. Thanks

PR #213, opened 2016-08-17
added inputName field to each dropped file in files array

gwenf:
Needed a way to differentiate between multiple react-dropzone components on the same page. Updated onDrop function. Allows this.props.name to be stored on the image file object as `inputName`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,10 @@
     "deps:update": "npm-check -u"
   },
   "lint-staged": {
-    "*.js": ["eslint:fix", "git:add"]
+    "*.js": [
+      "eslint:fix",
+      "git:add"
+    ]
   },
   "pre-commit": [
     "lint-staged"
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -96,6 +96,10 @@ class Dropzone extends React.Component {
 
     for (let i = 0; i < max; i++) {
       const file = droppedFiles[i];
+      
+      // Let's add inputName to file, so we can differentiate multiple dropzones
+      file.inputName = this.props.name ? this.props.name : '';
```
okonet: Wouldn't `file.inputName = this.props.name || '';` be better since `name` is always a String?

okonet:
Not sure how this is helping to differentiate between different dropzones/inputs. Can you provide an example? Wouldn't it be better to solve in the user-land with, say, list of file arrays?

Also, if you want me to merge this PR, please add tests for this change.

okonet:
Also, please fix ESLint errors: https://travis-ci.org/okonet/react-dropzone/builds/152982259#L457

PR #220, opened 2016-08-28
Allow style to be merged instead of being replaced

id0Sch:
let's say i want to increase height , i should be able to do so by setting only `height:500`
right now i need to actually go to the source code and copy the entire style section

hope you'll find that change good,
great job on the repo man! 
thanks!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -7,6 +7,14 @@ const supportMultiple = (typeof document !== 'undefined' && document && document
   'multiple' in document.createElement('input') :
   true;
 
+const defaultStyle = {
+  width: 200,
+  height: 200,
+  borderWidth: 2,
+  borderColor: '#666',
+  borderStyle: 'dashed',
+  borderRadius: 5
+};
 class Dropzone extends React.Component {
   constructor(props, context) {
     super(props, context);
@@ -167,15 +175,7 @@ class Dropzone extends React.Component {
       className += ' ' + rejectClassName;
     }
 
-    if (!className && !style && !activeStyle && !rejectStyle) {
-      style = {
-        width: 200,
-        height: 200,
-        borderWidth: 2,
-        borderColor: '#666',
-        borderStyle: 'dashed',
-        borderRadius: 5
-      };
+    if (!className && !activeStyle && !rejectStyle) {
       activeStyle = {
         borderStyle: 'solid',
         backgroundColor: '#eee'
@@ -189,16 +189,19 @@ class Dropzone extends React.Component {
     let appliedStyle;
     if (activeStyle && isDragActive) {
       appliedStyle = {
+        ...defaultStyle,
         ...style,
         ...activeStyle
       };
     } else if (rejectStyle && isDragReject) {
       appliedStyle = {
+        ...defaultStyle,
         ...style,
         ...rejectStyle
       };
     } else {
       appliedStyle = {
+        ...defaultStyle,
         ...style
       };
     }


okonet:
Hey! 

First of all, I'd like to thank you for your contribution. Can you explain how is that different from just passing the `style={{height: 500}}` as a prop? I can't see the difference now.

I like that `defaultStyle` got separated from `style`, though. This change, although, made the `let style` redundant. This is why the Travis CI build fails. Please fix linting errors and I'm good to merge.

BTW, did you do `npm install` after checking out the repo? Since in this case you should have gotten a pre-commit hook with linting.

id0Sch:
when i set height:500 as a prop the dashed border disappeared since my style replaced the default style.
from my experience with external react components, override is better than replace.
right now my component that wraps dropZone looks like that:
``` javascript
const dropZoneStyle = {
    width: '100%',
    height: 100,
    borderWidth: 2,
    borderColor: '#666',
    borderStyle: 'dashed',
    borderRadius: 5
};
const AddPackage = ({Upload:{file}, addFiles})=> {
    return (
        <Card>
            <CardText>
                {file ?
                    <div>
                        <Subheader>selected file</Subheader>
                        <KeyValueTable obj={_.pick(file, ['size', 'name'])} enableHrefs={false}/>

                    </div>
                    :
                    <Dropzone multiple={false} style={dropZoneStyle} onDrop={addFiles}>
                        <h4 style={{textAlign: 'center'}}>Click / Drop Apk file here</h4>
                    </Dropzone>
                }
            </CardText>
            <Divider/>
            <CardActions>
                <FlatButton label="clear"></FlatButton>
                <FlatButton primary={true} label="upload"></FlatButton>
            </CardActions>
        </Card>
    );
};
```
the
```
const dropZoneStyle = {
    width: '100%',
    height: 100,
    borderWidth: 2,
    borderColor: '#666',
    borderStyle: 'dashed',
    borderRadius: 5
};
````
with my PR is would be 
```
const dropZoneStyle = {
    width: '100%'
};
```

regarding the style var
should i just do
``` const {style} = rest ``` ?

okonet:
Thanks for the explanataion. Honestly, I don't think having implicit styles is better. Imagine the default styles were changed in the next version. With your approach, it will be a breaking change that will affect lots of users. I even think that no default styles should be in the library itself and I think I'll make this change in the next version (if I ever find time to release it).

Thoughts?

okonet:
Re: eslint — yes, this is what it says, no `let` for variables without reassign.

id0Sch:
well, you do have a point - I agree.
You could instead of removing the default style not apply it as a default but to export it maybe, so users can choose to depend on it or implement their own.
so we could do ```import {DropZone, defaultStyle} from 'dropzone'```
what do you think?

aboust eslint - i'll fix it when we'll decide if my PR is actually needed :)

okonet:
@id0Sch why would you prefer to depend on the lib's styles? There isn't much in there and I guess most people would like to style it their own they. But I'll give it some more thoughts.

id0Sch:
Well, in some cases you just want an out-of-the-box lib that just works without having to configure stuff.
In my case I wanted to use the default style but just change the height.
maybe to best solution is really to remove the default style and just have an example in the readme that shows the default style

okonet:
I've created #221 so let's discuss there.

id0Sch:
great, close this for now?

unbalanced:
Hey guys, 

I love to see components released with default look&feel. it easier to start using components that already has some default look.

The problem i see is that Inline-ing css in components makes it harder to reuse the component in different projects, since different projects usually means different sizes/colors/etc.. and some inline styles are just not possible to override (they are strong like !important)
Note, it might not be the case here, but many components have more than just one div and thus the inline styles on child elements are not reachable to override like in DropZone and this is where inline styles suck the most. 

We'll have a better developer-experience and a better overall-modularity by using className. 
If we'll let the component-creator to set his default lock&feel and the reusing-developer to easily override css rules if needed by extending the default class of every html element that the component is creating.

WDYT?

okonet:
I've decide to close it for now since this will require some more consideration and investigattions. Please see https://github.com/okonet/react-dropzone/issues/221 for the discussion.

PR #222, opened 2016-08-29
updated readme to include minSize and maxSize features

lemuelbarango:
Updated readme to include minSize and maxSize #202


okonet:
Awesome! Thanks!

PR #231, opened 2016-09-12
Add onclick event callback

josh18:
Does this look ok? Let me know if you need me to change anything.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -138,8 +138,12 @@ class Dropzone extends React.Component {
   }
 
   open() {
+    const clickEvent = new MouseEvent('click', {
+      bubbles: false
+    });
+
     this.fileInputEl.value = null;
-    this.fileInputEl.click();
+    this.fileInputEl.dispatchEvent(clickEvent);
   }
 
   render() {
--- a/src/index.js
+++ b/src/index.js
@@ -119,9 +119,13 @@ class Dropzone extends React.Component {
     }
   }
 
-  onClick() {
+  onClick(e) {
     if (!this.props.disableClick) {
       this.open();
+    }
+
+    if (this.props.onClick) {
+      this.props.onClick.call(this, e);
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -134,8 +138,12 @@ class Dropzone extends React.Component {
   }
 
   open() {
+    const clickEvent = new MouseEvent('click', {
+      bubbles: false
+    });
+
     this.fileInputEl.value = null;
-    this.fileInputEl.click();
+    this.fileInputEl.dispatchEvent(clickEvent);
```
okonet: Why is that needed?

src/index.js
```diff
@@ -119,10 +119,14 @@ class Dropzone extends React.Component {
...
   }
 
-  onClick() {
+  onClick(e) {
     if (!this.props.disableClick) {
       this.open();
     }
+
+    if (this.props.onClick) {
+      this.props.onClick.call(this, e);
```
okonet: Should probably also check if it is a `function` Here for better guarding.

okonet:
Tests are failing on CI.

src/index.js
josh18: I just copied the previous callbacks style, example :https://github.com/okonet/react-dropzone/blob/master/src/index.js#L53-L55

Also not sure if checking is a function is necessary due to it being checked in propTypes?

src/index.js
josh18: Otherwise the click bubbles up and triggers the DropZone click event again. Example: https://jsfiddle.net/8c13nd2y/

josh18:
Yeah I saw the tests are failing but I'm not sure why, [MouseEvent is a valid API](https://developer.mozilla.org/en/docs/Web/API/MouseEvent), maybe something needs to be added to `mocha-environment.js` to make it pass?

src/index.js
okonet: Probably. Leave it for now. This needs to be refactored for all places. FIY propTypes does not guard in runtime, that's why I think it's a good idea to check the type.

src/index.js
okonet: Why not just cancel bubbling in the event handler then? This looks like a hack to me.

okonet:
Yeah, we are using jsdom and it doesn't have all browser APIs out of the box. You have to mock it I guess.

src/index.js
josh18: You mean like adding `onClick: (e) => { e.stopPropagation(); }` here? https://github.com/okonet/react-dropzone/blob/master/src/index.js#L207-L214

okonet: Yes. This will do the same, isn't it?

josh18: Unfortunately not (that's what I tried first), the onClick that happens there is a synthetic React event so `e.stopPropagation` doesn't work on it. See: http://stackoverflow.com/a/24421834/1661462

okonet:
Can you please enable this https://github.com/blog/2247-improving-collaboration-with-forks so I can play around with it?

josh18:
It should already be enabled?

okonet:
Closing it in favor of https://github.com/okonet/react-dropzone/commit/33813573 which is already been released in https://github.com/okonet/react-dropzone/releases/tag/v3.8.0

PR #237, opened 2016-09-21
Issue-199 request for onCancel

anuragCES:
- Passes tests
- onDrop event gets called before body onfocus event, to resolve this added a timeout for 300ms.


src/index.js
```diff
@@ -16,6 +16,7 @@ class Dropzone extends React.Component {
     this.onDragLeave = this.onDragLeave.bind(this);
     this.onDragOver = this.onDragOver.bind(this);
     this.onDrop = this.onDrop.bind(this);
+    this.handleOnCancel = this.handleOnCancel.bind(this);
```
okonet: The naming is inconsistent with the rest of methods. Also, I can't figure out from the name what's for native event it represents.

src/index.js
```diff
@@ -24,6 +25,10 @@ class Dropzone extends React.Component {
 
   componentDidMount() {
     this.enterCounter = 0;
+    const { handleOnCancel } = this;
+    document.body.onfocus = (event) => {
```
okonet: You should properly attach the event listener to the DOM node using `addEventListener` method. Also, you forgot to remove the listener in `componentWillUnmount`

okonet: Also, instead of creating a clojure here `(event) => { handleOnCancel(event) }` you could just pass the  `this.handleOnCancel` as a listener since it has the same signature.

src/index.js
```diff
@@ -273,6 +290,7 @@ Dropzone.propTypes = {
 
   disablePreview: React.PropTypes.bool, // Enable/disable preview generation
   disableClick: React.PropTypes.bool, // Disallow clicking on the dropzone container to open file dialog
+  close: React.PropTypes.func,
```
okonet: What's this prop is about? You can't tell from the name what it is supposed to do.

src/index.js
```diff
@@ -125,6 +130,18 @@ class Dropzone extends React.Component {
...
   }
 
+  handleOnCancel() {
+    const { close } = this.props;
+    const { fileInputEl } = this;
+    setTimeout(() => {
+      const FileList = fileInputEl.files;
+      if (!FileList.length) {
+        document.body.onfocus = null;
+        close();
```
okonet: Since it isn't a required prop, this will throw an error if it is not specified.

src/index.js
```diff
@@ -125,6 +130,18 @@ class Dropzone extends React.Component {
     }
   }
 
+  handleOnCancel() {
+    const { close } = this.props;
+    const { fileInputEl } = this;
+    setTimeout(() => {
+      const FileList = fileInputEl.files;
+      if (!FileList.length) {
+        document.body.onfocus = null;
```
okonet: Same as with `addEventListener`, you should use `removeEventListener` here. What I don't get is, why are you removing it every time but adding just once?

anuragCES:
@okonet 

Thank you for your feedback.

I have changed the code as per your review. Also, I have added tests for possible scenarios. Kindly review and share your thoughts.

Although, as per your suggestion regarding adding an event listener. I was unable to add it on document.body or document. If you could help me it would be of great help.

Thank you.

akkie:
Is there something I can help here to get this merged? This would help me to use this component in a [redux friendly way](http://stackoverflow.com/questions/39954168/how-should-i-handle-os-specific-dialog-windows-in-a-redux-friendly-way).

.gitignore
```diff
@@ -6,7 +6,8 @@ logs
 node_modules
 
 # Generated files
-dist
```
okonet: Please add this back and remove generated files from the repository. They will be generated and put on npm during the build phase.

src/index.js
```diff
@@ -16,14 +16,23 @@ class Dropzone extends React.Component {
...
-      isDragActive: false
+      isDragActive: false,
+      isFileDialogActive: false
     };
   }
 
   componentDidMount() {
     this.enterCounter = 0;
+    // Tried implementing addEventListener, but didn't work out
+    document.body.onfocus = this.onFileDialogCancel;
```
okonet: That's weird. I would assume it can be the React that is messing around with the event listeners here... I could investigate further.

anuragCES:
@okonet 

Thanks 👍

anuragCES:
@okonet 

Fresh npm install do not have /dist directory in-turn causing an issue.

Kindly suggest.

okonet:
This is how it looks for me after `npm install react-dropzone`:

```
node_modules/react-dropzone
├── CHANGELOG.md
├── LICENSE
├── README.md
├── dist
│   ├── index.js
│   └── index.js.map
├── mocha-environment.js
├── package.json
├── src
│   ├── index.js
│   └── test.js
└── webpack.config.js
```

anuragCES:
@okonet 

It worked, It was an overlook from my end.

Thank you for helping out.

anuragCES:
@okonet 

onFileDialogCancel event invokes when I provide direct reference to src/index.js. However, it fails when used through /dist. Could you confirm the same?

Thank you.

okonet:
This wasn't released yet and I'll release it as part of 4.0 which can take some time since there are many changes in there.

akkie:
@anuragCES This feature doesn't work if the file dialog will be opened with the [open](https://github.com/okonet/react-dropzone/blob/master/src/index.js#L179) method, because the state `isFileDialogActive` will not be set.

anuragCES:
@akkie I'm using this.refs.dropzone.open() method to open the file dialog. Is this you're referring to?

Could you give a try referencing - "react-dropzone": "https://github.com/anuragCES/react-dropzone"
in you package.json?

Let me know if this works for you.

akkie:
There is also a problem with [this setState](https://github.com/okonet/react-dropzone/blob/master/src/index.js#L160) call. The second parameter passed to the `setState` function must be a callback. You pass the result of a function. Therefore the browser shows an error like:
```
invariant.js:38 Uncaught Invariant Violation: setState(...): Expected the last optional `callback` argument to be a function. Instead received: Object (keys: type, visible).
```

https://facebook.github.io/react/docs/component-api.html

@anuragCES Do you provide a fix for these both issues?

anuragCES:
@akkie 

Sure, I will raise a PR for the same.

anuragCES:
@okonet @akkie 

Here's the PR - https://github.com/okonet/react-dropzone/pull/253

PR #250, opened 2016-10-10
Separate accepted and rejected files

akkie:
Currently it's not possible to distinguish between accepted and rejected files. The `onDrop` and the `onDropAccepted` functions gets only called with the accepted files. The `onDropRejected` function on the other side gets only called with the rejected files. But sometimes it's useful to react to both states.

This pull request implements the feature in a backward compatible way. The `onDrop` function gets the rejected files passed as second parameter. The `onDropAccepted` and the `onDropRejected` functions behaves the same.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -93,7 +93,8 @@ class Dropzone extends React.Component {
 
     const droppedFiles = e.dataTransfer ? e.dataTransfer.files : e.target.files;
     const max = this.props.multiple ? droppedFiles.length : Math.min(droppedFiles.length, 1);
-    const files = [];
+    const acceptedFiles = [];
+    const rejectedFiles = [];
 
     for (let i = 0; i < max; i++) {
       const file = droppedFiles[i];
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -125,12 +131,16 @@ class Dropzone extends React.Component {
...
   }
 
-  allFilesMatchSize(files) {
-    return files.every(file => (file.size <= this.props.maxSize && file.size >= this.props.minSize));
+  fileMatchSize(file) {
+    return file.size <= this.props.maxSize && file.size >= this.props.minSize;
+  }
+
+  allFilesAccepted(files) {
+    return files.every(file => this.fileAccepted(file));
```
okonet: This can be re-written as `files.every(this.fileAccepted)`

okonet: No problem! That's why code review is great :)

akkie:
Yes, it can break code, because the `onDrop` method gets always called now. Previously it was only called if all files were accepted. So I think a major release would be the best. 

I think it may also be useful to pass the accepted files as second parameter to the `onDropRejected` function. So a user could display an error message like "2 files from 8 couldn't be uploaded". What do you think?

With this addition, every method gets now all files passed.
* `onDrop(passedFiles, rejectedFiles)` => Gets always called and is useful to handle custom behavior
* `onDropAccepted(files)` => Gets called if all files are accepted and is useful to handle the success state
* `onDropRejected(rejectedFiles, passedFiles)` => Gets called if any of the files will be rejected and is usefull to handle the failed state

akkie:
I've updated the pull request and I've changed only the style issue. Let me know if I should also implement the suggested `onDropRejected` behavior.

okonet:
@akkie I'm not sure we should change the signature of this method. In the described use case the `onDrop` method should be used to implement the custom logic. What we could do to make it simplier is to export filter function i.e. allow custom filters. See #104

okonet:
BTW for your example of showing that some files didn't get accepted, it's not even needed to export anything. So you could do:

```js
onDrop = (acceptedFiled, rejectedFiles) => {
    if (rejectedFiles.length) {
        // show message that some files were rejected
    }
    this.setState({
        files: acceptedFiles
    });
}
```

akkie:
Yes, you are right. I thought it would be useful for the sake of consistency. But I'm also fine with the current implementation. Anyway the filter option should be implemented into another pull request.

src/index.js
```diff
@@ -125,12 +131,16 @@ class Dropzone extends React.Component {
...
   }
 
-  allFilesMatchSize(files) {
-    return files.every(file => (file.size <= this.props.maxSize && file.size >= this.props.minSize));
+  fileMatchSize(file) {
+    return file.size <= this.props.maxSize && file.size >= this.props.minSize;
+  }
+
+  allFilesAccepted(files) {
+    return files.every(this.fileAccepted);
```
okonet: I've just noticed that https://github.com/okonet/react-dropzone/pull/250/files#diff-1fdf421c05c1140f6d71444ea2b27638R135 is using `this` but the method isn't bound to this class. Not sure if this part is covered by tests but it seems that `this.props` will be always `undefined` in this context. Or am I missing something?

akkie: The `fileAccepted` method is bound to the class and the `props` also. This works as expected it's also covered through the test.

akkie:
Great :+1:

macrozone:
This change is in the documentation, but there was no release on npm.

okonet:
I've released a new version: https://github.com/okonet/react-dropzone/releases/tag/3.7.0

PR #253, opened 2016-10-13
Issue-199 : Refactored open method

anuragCES:


anuragCES:
This PR is referenced from the conversations at - https://github.com/okonet/react-dropzone/pull/237

src/index.js
```diff
@@ -178,6 +176,7 @@ class Dropzone extends React.Component {
   }
 
   open() {
+    this.setState({ isFileDialogActive: true });
```
akkie: You could also write:
```js
this.setState({ isFileDialogActive: true }, () => {
  this.fileInputEl.value = null;
  this.fileInputEl.click();
});
```

I'm not sure which is the better approach

anuragCES: @okonet @akkie 

Thank you for your feedback.

```
open() {
    const { fileInputEl } = this;
    this.setState({ isFileDialogActive: true }, () => {
      fileInputEl.value = null;
      fileInputEl.click();
    });
  }
```

This causes infinite loop since the `fileInputEl.click()` make the open method to called again.

Shall I change the `isFileDialogActive` to be class variable rather than a state. I can change the usage from `this.state.isFileDialogActive` to `this.isFileDialogActive` everywhere.

Need suggestions?

okonet: Yes, that makes more sense since we don't actually want to re-render when changing the value. Just to track it.

anuragCES:
@okonet 

Please go ahead with the code review.

Thank you.

okonet:
@anuragCES let's fix the failing test before I do please.

anuragCES:
@okonet 

It passes test cases now.

stipsan:
Would be neat with a new release as we're facing this issue after moving to react-router@v4 😄

okonet:
Let me cut the release.

stipsan:
Thanks!

PR #257, opened 2016-10-19
Add rejectionReason property to rejected files

rrhvella:
The rationale behind this is that I want to inform the user about the reason that a particular file was rejected. With this property I can listen to onDropRejected or onDrop, add the rejected files to my list, and display a tooltip giving the reason as translated text.

rejectionReason is only defined on rejected files. And has one of three possible values.

1 = InvalidFileType
2 = BelowMinSize
3 = AboveMaxSize

These values are listed on the RejectionReason enum which is a static property on the Dropzone class.

`
Dropzone.RejectionReason = {
  InvalidFileType: 1,
  BelowMinSize: 2,
  AboveMaxSize: 3
};
`

Thank you for the good work and I look forward to your feedback.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -108,14 +108,26 @@ class Dropzone extends React.Component {
 
     for (let i = 0; i < max; i++) {
       const file = droppedFiles[i];
+
       // We might want to disable the preview creation to support big files
       if (!this.props.disablePreview) {
         file.preview = window.URL.createObjectURL(file);
       }
 
-      if (this.fileAccepted(file) && this.fileMatchSize(file)) {
+      let rejectionReason = null;
+
+      if (!this.fileAccepted(file)) {
+        rejectionReason = Dropzone.RejectionReason.InvalidFileType;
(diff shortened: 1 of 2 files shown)


kandros:
LGTM

okonet:
First of all thanks for your contribution and sorry, I dropped the ball on it. I'm still not sure the design of this feature is complete enough. For example, when trying to drop several files, each of this files can be rejected with a different reason. Question: what reason will be passed in this case?

I was thinking about implementing custom validators and would like to discuss it in a separate issue here: #321

kandros:
> For example, when trying to drop several files, each of this files can be rejected with a different reason. Question: what reason will be passed in this case

as shown here https://github.com/okonet/react-dropzone/pull/257/files#diff-1fdf421c05c1140f6d71444ea2b27638R130 the reason of failure gets injected to the file itself

so in `this.props.onDropRejected.call(this, rejectedFiles, e)` the rejected files array have reasons

okonet:
Closing based on #321

Issue #252, opened 2016-10-12
<Dropzone> cause onClick to fire twice on parent <div>


aaronbeall:
I ran into an issue with click handlers and drag handlers being fired twice on a `<div>`, and found that it happens when I have `<Dropzone>` as a child of the `<div>`.

Before `<Dropzone>`:

``` jsx
<div onClick={e => console.log("click", e)}>
</div>
```

Clicking the `<div>` results in:
![image](https://cloud.githubusercontent.com/assets/2973900/19330301/9d706e0a-90ab-11e6-8219-6146608b7fbf.png)

After `<Dropzone>`:

``` jsx
<div onClick={e => console.log("click", e)}>
  <Dropzone className="upload-drop-area"
            activeClassName="upload-drop-area-accept"
            rejectClassName="upload-drop-area-reject"
            multiple={false}
            onDragEnter={this.onDragEnter}
            onDragLeave={this.onDragLeave}
            onDropAccepted={this.onDropAccepted}>
  </Dropzone>
</div>
```

Clicking the `<div>` results in two events:
![image](https://cloud.githubusercontent.com/assets/2973900/19330370/04e352a0-90ac-11e6-8cd1-3a6511e365d1.png)

I've never seen this kind of thing in React before. Any idea what's going on?

aaronbeall:
A few more observations:

* If I replace the `<Dropzone>` with equivalent content `<div><input/></div>` the problem doesn't exist. Only when I use `<Dropzone>` does it happen.

* The problem happens for `onClick` but not `onMouseOver`.

okonet:
`Dropzone` handles clicks and it will bubble the event up to its parents. 

https://github.com/okonet/react-dropzone/blob/master/src/index.js#L276

You should prevent the event going down the tree in your handler if you want to prevent this behavior.

aaronbeall:
I'm not sure I understand. Bubbling shouldn't result in multiple events.... you're saying that because Dropzone has an `onClick` handler that causes two callbacks to fire on the parent?

aaronbeall:
I think this is the problem:

https://github.com/okonet/react-dropzone/blob/master/src/index.js#L182

You are calling `click()` on the input from an `onClick` handler, the input click event bubbles so you end up with two `onClick` callbacks getting fired.

okonet:
im not quite sure about the behavior. Can you create an isolated example with a vanilla JS and DOM? We probably should stop propagation as you said. Mind creating a PR?

[...]

PR #273, opened 2016-11-09
Fixed double `onClick` event on parent element

aaronbeall:
Fixes https://github.com/okonet/react-dropzone/issues/252

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -138,7 +138,7 @@ class Dropzone extends React.Component {
       this.open();
     }
   }
-  
+
   onInputClick(e) {
     e.stopPropagation();
   }
--- a/src/index.js
+++ b/src/index.js
@@ -17,6 +17,7 @@ class Dropzone extends React.Component {
     this.onDragOver = this.onDragOver.bind(this);
     this.onDrop = this.onDrop.bind(this);
     this.onFileDialogCancel = this.onFileDialogCancel.bind(this);
+    this.onInputClick = this.onInputClick.bind(this);
     this.fileAccepted = this.fileAccepted.bind(this);
     this.isFileDialogActive = false;
     this.state = {
@@ -136,6 +137,10 @@ class Dropzone extends React.Component {
     if (!this.props.disableClick) {
       this.open();
     }
+  }
+  
+  onInputClick(e) {
+    e.stopPropagation();
   }
 
   onFileDialogCancel() {
@@ -284,6 +289,7 @@ class Dropzone extends React.Component {
         <input
           {...inputProps/* expand user provided inputProps first so inputAttributes override them */}
           {...inputAttributes}
+          onClick={this.onInputClick}
         />
       </div>
     );


okonet:
Thanks for the PR!

Two things before I'll merge it:

1. The build fails due to some eslint errors. Please fix those.
1. I think we'll need a test for this new code. It should cover the case you've described in #252. Can you please add it?

BTW please do `npm install` — it will install `lint-staged` that should prevent eslint-broken commits.

aaronbeall:
I ran `npm install` and `npm build` and I don't see the errors... what do I need to run to see these errors?

aaronbeall:
Ah I see it running `npm run eslint:src`

okonet:
You should have seen these before commit. This project has an integrated pre-commit hook.

aaronbeall:
I tried to add this test:

```
      const onClickSpy = spy();

      const component = TestUtils.renderIntoDocument(
        <div onClick={onClickSpy}>
          <Dropzone className="dropzone-content" />
        </div>
      );

      const content = TestUtils.findRenderedDOMComponentWithClass(component, 'dropzone-content');
      TestUtils.Simulate.click(content);
      expect(onClickSpy.callCount).to.equal(1);
```

However this gives me `Invariant Violation: findAllInRenderedTree(...): instance must be a composite component` apparently because [you can't use this with a DOM element](https://github.com/facebook/react/issues/4692).

So I tried this:

```
      const onClickSpy = spy();

      const WrappedDropzone = () => (
        <div onClick={onClickSpy}>
          <Dropzone className="dropzone-content" />
        </div>
      );

      const component = TestUtils.renderIntoDocument(
        <WrappedDropzone />
      );

      const content = TestUtils.findRenderedDOMComponentWithClass(component, 'dropzone-content');
      TestUtils.Simulate.click(content);
      expect(onClickSpy.callCount).to.equal(1);
```

Which throws: `Error: Did not find exactly one match (found: 0) for class:dropzone-content`. 

Any tips?

aaronbeall:
Yes I should have seen the ESLint pre-commit errors but I edited it straight on Github after messing around with things locally... which was silly. :)

okonet:
@aaronbeall any updates on this PR? Do you think you will be able to add tests for this case?

aaronbeall:
I got stuck on that test in my last reply and haven't been able to circle back to figure out how to workaround the `TestUtils` limitation... my quick research made it sound like you aren't expected to test a wrapping element with `TestUtils`. 😖

okonet:
> my quick research made it sound like you aren't expected to test a wrapping element with TestUtils

I'm not sure what you mean by that. Can you elaborate please?

aaronbeall:
This: https://github.com/facebook/react/issues/4692

I'm not sure that's the issue but that was as far as I got...

okonet:
@aaronbeall I was able to repro the case in tests: https://github.com/okonet/react-dropzone/pull/309/files#diff-3d2fceb7429c5721d178682a76d75809R514

I'm skipping the test for now but it is probably a good idea to move it to this branch and land as a part of this PR. Let me finish my work and let's update this one.

okonet:
This turned out to be a bit trickier and required me to add `onClick` callback support so I'm fixing this in #312. Thanks for pointing out and your effort!

PR #301, opened 2016-12-08
hey, fix for FF 

tcat:
hey, it doesnt work in FF, with this fix it should(tested in Chrome and FF)
Main problem was that in FF doesn't work activeClassName for accepted files. This fix resolves this problem


okonet:
Thanks! I'm implementing it in https://github.com/okonet/react-dropzone/pull/309

PR #303, opened 2016-12-13
Fullscreen support for dropzone.

matanasa:


okonet:
@matanasa I'm not sure what you're doing but you shouldn't change version and add dist files to this repo. Also, before you submit your PR let's discuss it. There are a few open issues about full screen mode but I'm still not convinced it should be part of the lib.

okonet:
See https://gist.github.com/wvance/c052a57654ea943edee113a180598ab8 for example

PR #322, opened 2017-01-11
Remove filetype check for preview

zenners:
I used to be able to preview PDFs. This PR seemed simple enough haha

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -102,7 +102,7 @@ class Dropzone extends React.Component {
     this.isFileDialogActive = false;
 
     fileList.forEach((file) => {
-      if (!disablePreview && accepts(file, 'image/*')) {
+      if (!disablePreview) {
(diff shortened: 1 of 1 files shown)


okonet:
Hmm, it should fail the test but it didn't 🤔

okonet:
This is embarrassing but I forgot to remove `.skip` from the test: https://github.com/okonet/react-dropzone/blob/master/src/index.spec.js#L541

Can you please update the spec file as well so it corresponds this behavior. Also, should this be considered a breaking change?

mkuipers:
>Can you please update the spec file as well so it corresponds this behavior. Also, should this be considered a breaking change?

Doesn't really seem like a breaking change - seems like a bugfix for something that we'd expected to have been caught by the spec failure.

okonet:
I'm closing this in favor of #332

PR #329, opened 2017-01-13
Allow string array's for accept argument

smeijer:
As [`attr-accept`][1] can handle array's as well; and this library simply passes the argument trough; I don't see any reason to leave this one out of the `PropTypes`. 

```js
accept({
    name: 'my file.json',
    type: 'application/json'
}, 'application/json,video/*') // => true

accept({
    name: 'my file.json',
    type: 'application/json'
}, ['application/json', 'video/*']) // => true
```

[1]: https://github.com/okonet/attr-accept#usage

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -331,7 +331,10 @@ Dropzone.propTypes = {
 
   inputProps: React.PropTypes.object, // Pass additional attributes to the <input type="file"/> tag
   multiple: React.PropTypes.bool, // Allow dropping multiple files
-  accept: React.PropTypes.string, // Allow specific types of files. See https://github.com/okonet/attr-accept for more information
+  accept: React.PropTypes.oneOfType([  // Allow specific types of files. See https://github.com/okonet/attr-accept for more information
+    React.PropTypes.string,
+    React.PropTypes.arrayOf(React.PropTypes.string)
+  ]),
   name: React.PropTypes.string, // name attribute for the input tag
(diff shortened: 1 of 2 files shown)


okonet:
The reason is that it won't be supported by `input` element since it's not a standard AFAIK

okonet:
@smeijer can you make a research and check if arrays will work with the `input`

smeijer:
@okonet , how about adding a test? React seems to transform the array to a string itself.

okonet:
Hmm, this seems very fragile to me. If React changes how this behaves this will break lots of code.

Why not just use strings as the standard prescribes? Seems very simple to just use `.join(',')` in your code. 

Thanks for your effort but I really like this to stay in line with the standard which doesn't allow arrays.

smeijer:
With a [commit like this](https://github.com/smeijer/react-dropzone/commit/fff5a8d82a9d781e74845db5e55aa9953f246614); you don't even need to depend on React parsing the attribute. 

Although, I don't understand why you call it fragile; as it's just as "fragile" as depending on the React SyntheticEvent, I do understand you don't want to support this. 

I'll just use a wrapper for this functionality:

```js
// dropzone.js
import Dropzone from 'react-dropzone';

export default ({ accept, ...props }) =>
  <Dropzone accept={Array.isArray(accept) ? accept.join(',') : accept} {...props} />;
```

PR #335, opened 2017-01-25
fix(onClick): don't swallow click events when disableClick is true

frankwallis:
Scenario that I have is that list items are drop targets but users can also select items in the list by clicking on them. In 3.9.1 click events are not being received by the list items as they are swallowed by the dropzone.

Let me know if you would like anything changed, thanks.


okonet:
Awesome! Thanks!

Issue #161, opened 2016-04-05
Remove display none from input element


vladshcherbin:
Since it is hidden, we can't use Tab to select it and choose a file.

If is remains visible, but has style of width: 0, height: 0 - we are able to choose this element.

okonet:
PR?

VladShcherbin:
@okonet sure, will do today

okonet:
Possible related to #180 

@VladShcherbin what about the PR? Still gonna do it?

VladShcherbin:
@okonet hey, I've tried to do it, but we need to show the active state somehow with css (as you press tab  key and want to see it becomes active).

We can use `+` selector, but this way we need to make sure `input` element is placed before the styled element.

okonet:
I'm open to any solution at this point. I also think that current implementation prevents using Dropzone inside of `form` element with `submit` handler, so I'd say, it's worth trying. I like the idea BTW and glad to help.

arjshiv:
Naive question:, but can we use`opacity: 0` instead of `display: none` ? That should create an element on the stacking context which should be clickable and tabbable http://stackoverflow.com/a/742324

VladShcherbin:
@arjshiv you can set the opacity, but the item (input element) will produce white space inside the dropbox element (width + height of the element).

But it's not the main problem - you need to show the user that the dropbox element gets focused. To do this - you can add a class when input is focused or use css selector.

VladShcherbin:
Probably an easy solution can be smth like this:

```js
const dropzoneClassName = classNames('dropzone', { focused: this.state.isFocused })

<Dropzone className={dropzoneClassName}>
  <input onFocus={this.handleInputFocus} />
</Dropzone>
```

And the input element can be set to 0 / 0 (width / height) and 0 opacity.

arjshiv:
@VladShcherbin Thanks - that's helpful. I'll try and see if I can take a crack at it later this week if you and @okonet are open to it

arjshiv:
@VladShcherbin yeah that's exactly the kind of thing I had in mind

VladShcherbin:
@arjshiv sure, go for it. I tried some time ago, but unfortunately didn't end up with the tests setup.

Would be great if you'll try to solve it ;)

VladShcherbin:
[...]

PR #336, opened 2017-01-26
Try fix keyboard accessibility

forabi:
Currently, the markup produced by `react-dropzone` is not accessible by keyboard navigation. The root element is a `div` and the `input` is hidden so it cannot be focused.

See also related discussion in #161 

To solve this problem, a `button` element should be used instead of a `div` at the root, and `tabIndex` should be allowed as a prop, this makes the dropzone focusable and allows the browser to trigger _a synthetic click_ when the user presses `Enter`.

## Why not `aria-*` markup?
Unfortunately, using `aria-*` markup does not cause the browser to trigger `click` handlers with the keyboard, so pressing `Enter` won't open the file picker, even when `onClick` is set. Here is an excellent video explaining why `button` is a better approach: https://www.youtube.com/watch?v=CZGqnp06DnI

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -278,7 +278,7 @@ class Dropzone extends React.Component {
     customProps.forEach(prop => delete divProps[prop]);
 
     return (
-      <div
+      <button
         className={className}
         style={appliedStyle}
         {...divProps/* expand user provided props first so event handlers are never overridden */}
@@ -294,7 +294,7 @@ class Dropzone extends React.Component {
           {...inputProps/* expand user provided inputProps first so inputAttributes override them */}
           {...inputAttributes}
         />
-      </div>
+      </button>
     );
   }
 }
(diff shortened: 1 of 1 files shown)


okonet:
First of all, thanks for you contribution. Please add a description of why you've created this PR.

Also from what I can see this is gonna break lots of UIs because of how `button` is styled in the browsers. Do you think we could do this with `aria-` tags somehow without changing the markup?

forabi:
@okonet I added a description of the changes and explained why `aria-*` does not solve the problem.

Breaking UIs is still concern though, can we "reset" the button style somehow so it looks exactly like a `div`?

okonet:
@forabi I think you could reset, but AFAIK there is no _really_ reliable way of doing this. But I agree accessibility is an issue now. Just wondering if there are more backward compatible solutions.

okonet:
Just FYI according to: http://caniuse.com/#feat=dragndrop

> In Firefox, the dragstart event does not fire on button elements. This effectively disables drag and drop for button elements.

forabi:
Turns out it's not a good idea. I had a case where I needed to place a button inside a dropzone and nested buttons are not valid HTML.

emily-plummer:
Hey @forabi happy to collaborate on this - I'd like to use dropzone as well but need it to be keyboard accessible. Can you explain where you are stuck with this PR?

holloway:
The `<input type=file>` is only `display:none` for visual reasons, right? So why not just put it in a container div whose `width:0,height:0,overflow:hidden`? That way it would still be keyboard navigable.

PR #338, opened 2017-01-30
Prevent dragging/dropping DOM elements from doing anything

### LINKED ISSUE #255: Suggestion: Option to prevent dragging and dropping elements on the page onto the dropzone
In the meantime, here is my solution for doing this:

```
onDragEnter={function( event ) {
    const dt = event.dataTransfer;
    if ( !(
        dt.types &&
        ( dt.types.indexOf ? dt.types.indexOf( "Files" ) !== -1 : dt.types.contains( "Files" ) )
    ) ) {
        this.setState( { isDragActive: false } );
    }
}}
```

```
onDrop={files => {
    if ( files.length === 0 ) {
        return;
    }

    // handle files dropped here
}}
```

buob:
Fixes #255

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -44,6 +44,14 @@ class Dropzone extends React.Component {
   onDragEnter(e) {
     e.preventDefault();
 
+    const dt = e.dataTransfer;
+
+    // If the object isn't a file (dom element, link, etc.)
+    if (!(dt.types && (dt.types.indexOf ? dt.types.indexOf('Files') !== -1 : dt.types.contains('Files')))) {
+      this.setState({ isDragActive: false });
+      return;
+    }
+
     // Count the dropzone and any children that are entered.
     ++this.enterCounter;
 
@@ -100,6 +108,11 @@ class Dropzone extends React.Component {
 
     // Stop default browser behavior
     e.preventDefault();
+
+    // If there are no files (i.e. if you drag a DOM element or image from within the browser)
+    if (fileList.length === 0) {
+      return;
+    }
 
     // Reset the counter along with the drag on a drop.
     this.enterCounter = 0;
--- a/src/index.spec.js
+++ b/src/index.spec.js
@@ -223,7 +223,7 @@ describe('Dropzone', () => {
             case 'dataTransfer':
               throw new Error('IE does not support rrror');
             default:
-              return function () {};
+              return function noop() {};
           }
         }
       });
--- a/src/index.spec.js
+++ b/src/index.spec.js
@@ -643,14 +643,37 @@ describe('Dropzone', () => {
 
   });
 
-  describe('behavior', () => {
-    it('does not throw an error when html is dropped instead of files and multiple is false', () => {
-      const dropzone = mount(
-        <Dropzone multiple={false} />
-      );
+  describe('edge cases', () => {
+    describe('when a DOM element is dragged/dropped instead of files', () => {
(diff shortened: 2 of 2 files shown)


buob:
Not sure why the PR isn't linked in the close link, but this fixes https://github.com/okonet/react-dropzone/issues/255

buob:
Also, not sure why codecov thinks I'm losing coverage, I'm adding quite a few more LOC of tests than actual code. 😞

buob:
@okonet Is there any update on this?  Is there something I need to change to make this a merge candidate?  I assumed at the time of submitting codecov was acting up, but if it's an actual problem I have, let me know 👍.

okonet:
I've received a second PR that should fix this issue: https://github.com/okonet/react-dropzone/pull/340 and from looking at the code I prefer it since it introduces much less new code. What are your thoughts? Does the second one missing some edge-cases that aren't covered in tests?

src/index.js
```diff
@@ -101,6 +109,11 @@ class Dropzone extends React.Component {
     // Stop default browser behavior
     e.preventDefault();
 
+    // If there are no files (i.e. if you drag a DOM element or image from within the browser)
+    if (fileList.length === 0) {
+      return;
```
okonet: I feel like we still want to reset `this.enterCounter` etc in this case.

src/index.js
```diff
@@ -44,6 +44,14 @@ class Dropzone extends React.Component {
   onDragEnter(e) {
     e.preventDefault();
 
+    const dt = e.dataTransfer;
+
+    // If the object isn't a file (dom element, link, etc.)
+    if (!(dt.types && (dt.types.indexOf ? dt.types.indexOf('Files') !== -1 : dt.types.contains('Files')))) {
```
okonet: `.contains` isn't implemented in most of the browsers yet AFAIK. I'd rather had this extracted to a util function with separate tests.

buob:
Hi @okonet!  My impetus for opening this PR was actually more that in Safari if you drag a DOM element (i.e. an image on the page) over the target, it triggers the drag active class.  The other PR doesn't address that.

I'm happy to investigate the other approach for the dropping portion though, it certainly is cleaner. 👍

okonet:
Yes, can you please add a description of which exact problems it is solving?

src/index.spec.js
```diff
@@ -636,14 +643,37 @@ describe('Dropzone', () => {
...
+        const dropzone = mount(
+          <Dropzone onDragEnter={onDragEnter} />
+        );
+
+        dropzone.simulate('dragEnter', { dataTransfer: { files: [], types: ['text/uri-list'] } });
+        expect(dropzone.state().isDragActive).toEqual(false);
+        expect(onDragEnter.called).not.toBeTruthy();
+      });
+
+      it.only('does not drop the items', () => {
```
masylum: ☝️

seberik:
@okonet Why did this become closed? I would be happy to continue working on a fix for this issue as we solved it with a onDragStart workaround. The problem is that we have a html5-drag-and-drop component inside of the dropzone component to allow images to be uploaded and sorted. When users sorts the uploaded images react-dropzone calls onDropRejected with a dataTransferItem of type "application/json".

rkmarks:
@seberik Correct, the issue was closed due to inactivity as mentioned by the stale bot. You can continue to work on your fix and make a PR whenever you can :)

PR #340, opened 2017-01-31
Ignore Drag'n'drop elements that are not files

afgomez:
Fixes an issue where we combine `Dropzone` with other
elements that handle Drag'n'drop, like [react-sortable].

When combining these libraries, the `onDrop` event of
`Dropzone` gets triggered and tried to make a preview
of something that was not a file, throwing an exception.

[react-sortable]: https://github.com/cheton/react-sortable

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -106,7 +106,7 @@ class Dropzone extends React.Component {
     this.isFileDialogActive = false;
 
     fileList.forEach((file) => {
-      if (file instanceof window.DataTransferItem) {
+      if (!(file instanceof window.File)) {
         return;
       }
 
--- a/src/index.js
+++ b/src/index.js
@@ -106,6 +106,10 @@ class Dropzone extends React.Component {
     this.isFileDialogActive = false;
 
     fileList.forEach((file) => {
+      if (file instanceof window.DataTransferItem) {
+        return;
+      }
+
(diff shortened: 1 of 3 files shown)


afgomez:
Mmm, I just checked, and I think this is a duplicate of #338

testSetup.js
```diff
@@ -5,3 +5,8 @@ global.window.URL = {
     return 'data://' + arg.name;
   }
 };
+
+global.window.DataTransferItem = function DataTransferItem(arg) {
```
okonet: It doesn't look like you need to expose this globally. Why not just use this function in test?

src/index.js
```diff
@@ -106,6 +106,10 @@ class Dropzone extends React.Component {
     this.isFileDialogActive = false;
 
     fileList.forEach((file) => {
+      if (!(file instanceof window.File)) {
```
okonet: What about the case there `window.File` is `undefined`? Also, this will probably break SSR implementations.

oyeanuj:
@afgomez @okonet Just checking to see the status of the PR? Do you see it being merged soon?

okonet:
Seems stalled to me :/

afgomez:
Sorry, I don't have time to look into it :(

I'll close the PR

Issue #204, opened 2016-08-03
Active state passed on as prop to children


aymericbouzy:
I would like to style the children according to the active state, not only the surrounding component.
One way to do this would be to provide a `renderChild` prop, which would be a function taking the boolean `active` as argument.
Maybe there is some other way I didn't think of?

okonet:
What you want is just passing the active state flag down the tree as a prop. PR?

aymericbouzy:
What you suggest makes more sense. I'll look into this, I'll keep you posted.

aymericbouzy:
Now that I have more experience with React, it seems to me that you cannot pass them as prop : children are rendered before Dropzone renders, it's too late to pass props to them. You could rather use something like context, but it doesn't seem designed to do that. 

`react-virtualized` uses a nice syntax I think. In this situation, it would allow to have this situation : 

```jsx
<Dropzone onDrop={this.onDrop}>
  {(active) => {
    if (active) {
      return <ActiveDropzoneContent />
    } else {
      return <InactiveDropzoneContent />
    }
  }}
</Dropzon />
```

I find it very readable, and not very difficult to implement. I can do a PR if you approve allowing such syntax (on top of the current one).

okonet:
It's totally possible to extend childrens props. I also like function syntax. It could do both I think. Please feel free to create a PR and let's discuss in it.

Issue #221, opened 2016-08-29
Should contain default styles or it should be up to user?


okonet:
I feel like if the dropzone component was unstyled by default, it would make it clear that it should be done in the user-land and no false expectation would be there.

On the other hand, integrating the library for the first time would require some more work without clear visual clues. This can be frustrating for the first-time users.

What styling strategy are you using in your project? Inline styles or passing class names? Should it support Radium or other styling libs?

After looking at #220, I'm not sure if the default styles should ever be the part of the library. If you're using it, please help me make a right decision.

okonet:
See a discussion also here: https://github.com/okonet/react-dropzone/pull/220#issuecomment-243050289

arjshiv:
@okonet I just started using this. We are using inline styles, and I thought that the default styling (with the dashed border) was very helpful in the initial integration. From a UI perspective, the one thing that would make it even easier is if it has a default `activeStyle` when you drag something in.  People with larger apps will probably override it anyway, but default active styles will make it even more of a plug'n'play experience for devs who want to get it done quickly.

I'm happy to help where required on this. Great project, would love to contribute!

josh18:
Maybe just add a bool property `defaultStyles` to turn the styles off / on? Of course if they have it turned on and add in their own styles then their own styles would take priority. I personally would have it off by default and then just make it really obvious in the docs that it is an option.

ajwagner777:
I hate to bring up a potentially old issue, but this issue is an important one to me (and possibly a deal breaker).  We are building a rather large app, and I would really prefer to keep the styling consistent with the rest of the app's branding.  I would propose the following:

- Add a class to the container that would allow styling hooks for the different states (default, drag over, invalid, others?)
- Add a prop that can conditionally remove the inlined styles
- Add some documentation to README 

This preserves current functionality for current users, and allows for much better styling.  I am happy to submit a PR to this effect.  Just wanted to make sure this was okay before putting in the work.

okonet:
@ajwagner777 the issue isn't old and still actual. Let me answer your questions:

> Add a class to the container that would allow styling hooks for the different states (default, drag over, invalid, others?)

There is a way of styling the dropzone the way you want. There are actually a few several ways: you can use `className` or inline styles. So, if you're using a `className` it will ignore inline styles completely. There is no documentation on that but if you look at `propTypes` here: https://github.com/okonet/react-dropzone/blob/master/src/index.js#L267-L272 you'll find all of them.

[...]

PR #342, opened 2017-02-02
Pass drag state to children

aymericbouzy:
**Work in progress**

PR related to issue #204 
If you approve the changes, I can suggest documentation for the feature.

With regards to what you said in the issue, I still don't understand how you can pass props to already _rendered_ children, but I'd love to see this!


okonet:
Hey @aymericbouzy! Thanks for the PR. 

You can see here how to extend props with the state here: https://github.com/okonet/react-container-dimensions/blob/master/src/index.js#L54 Mind doing something like this?

Also I'm wondering if it's a good idea to pass the whole `this.state` down the tree since if it will be extended or changed at some time point, it might break someone else's code. Probably this method act as proxy and map internal state to some generic argument names? Or at least we should probably explicitly filter out arguments we'd like to pass. What do you think?

aymericbouzy:
oh I didn't know about `React.cloneElement`, sure I can do something like this :) Thanks for pointing that out! Should I discard the function altogether, or provide both methods?

Totally makes sense to provide only the relevant parts of the state, but I wasn't sure which and how, that's why I suggested it this way ;) Should I rename the properties `active` and `rejecting`, or  should I keep the same names?

src/index.js
```diff
@@ -289,7 +298,7 @@ class Dropzone extends React.Component {
         onDragLeave={this.onDragLeave}
         onDrop={this.onDrop}
       >
-        {this.props.children}
+        {this.renderChildren()}
```
okonet: Should we pass children to this method and make it static?

src/index.js
```diff
@@ -184,6 +185,14 @@ class Dropzone extends React.Component {
     this.fileInputEl.click();
   }
 
+  renderChildren() {
+    let { children } = this.props;
+    if (typeof children === 'function') {
+      children = children(this.state);
```
okonet: You can `return` here instead of mutating the value.

src/index.js
```diff
@@ -184,6 +185,14 @@ class Dropzone extends React.Component {
     this.fileInputEl.click();
   }
 
+  renderChildren() {
+    let { children } = this.props;
```
okonet: Let's pass `children` to this function and remove `this` completely. In this case this method can be static.

src/index.js
aymericbouzy: Can you explain "making it static"?

aymericbouzy: Is that what you mean? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static

okonet: If you're not using `this` in the method, you can extract it as a function outside of the class. Or you could use `static` keyword.

aymericbouzy: Ok. I can do both, tell which you prefer ;) In my opinion, `static` makes more sense (that way it is scoped, etc), but it's an advanced feature of Javascript, so it makes it less readable for beginners. I'll do whichever you think is best.

okonet: Let's try `static` for now. This can always be refactored.

okonet:
> Totally makes sense to provide only the relevant parts of the state, but I wasn't sure which and how, that's why I suggested it this way ;) Should I rename the properties active and rejecting, or should I keep the same names?

I'd keep same names for now on. Also, what in your opinion should be passed down?

src/index.js
```diff
@@ -9,11 +9,17 @@ const supportMultiple = (typeof document !== 'undefined' && document && document
...
+    }
     let renderedChildren = children;
+    const props = { isDragActive, isDragReject };
     if (typeof children === 'function') {
-      renderedChildren = children(this.state);
+      renderedChildren = children(props);
     }
-    return React.cloneElement(renderedChildren, { isDragActive, isDragReject });
+    // a warning gets printed because we are giving unkown props to some children (<p>some content</p> for instance)
+    renderedChildren = React.cloneElement(renderedChildren, props);
```
aymericbouzy: In my opinion, doing this is bad. It may break some people's code (if they had components with `isDragActive` prop for instance). I think we should remove this step.

okonet: So you proposing to only use function and not `React.cloneElement`?

src/index.js
```diff
@@ -8,6 +8,20 @@ const supportMultiple = (typeof document !== 'undefined' && document && document
   true;
 
 class Dropzone extends React.Component {
+  static renderChildren(children, isDragActive, isDragReject) {
+    if (!children) {
+      return children;
+    }
+    let renderedChildren = children;
```
okonet: Why do you need this variable? It's an additional allocation that isn't needed. You can just return from `if` and in the end of this method.

src/index.js
```diff
@@ -8,6 +8,20 @@ const supportMultiple = (typeof document !== 'undefined' && document && document
...
+    const props = { isDragActive, isDragReject };
+    if (typeof children === 'function') {
+      renderedChildren = children(props);
+    }
+    // a warning gets printed because we are giving unkown props to some children (<p>some content</p> for instance)
+    renderedChildren = React.cloneElement(renderedChildren, props);
+    return renderedChildren;
```
okonet: `return React.cloneElement(renderedChildren, props);`

src/index.js
```diff
@@ -8,6 +8,20 @@ const supportMultiple = (typeof document !== 'undefined' && document && document
...
 
 class Dropzone extends React.Component {
+  static renderChildren(children, isDragActive, isDragReject) {
+    if (!children) {
+      return children;
+    }
+    let renderedChildren = children;
+    const props = { isDragActive, isDragReject };
+    if (typeof children === 'function') {
+      renderedChildren = children(props);
```
okonet: `return children(props);`

src/index.js
```diff
@@ -8,6 +8,20 @@ const supportMultiple = (typeof document !== 'undefined' && document && document
   true;
 
 class Dropzone extends React.Component {
+  static renderChildren(children, isDragActive, isDragReject) {
+    if (!children) {
+      return children;
+    }
+    let renderedChildren = children;
+    const props = { isDragActive, isDragReject };
```
okonet: Let's rename `props` to something with more context. Right now it's easy to confuse with `this.props`

src/index.js
```diff
@@ -8,6 +8,20 @@ const supportMultiple = (typeof document !== 'undefined' && document && document
   true;
 
 class Dropzone extends React.Component {
+  static renderChildren(children, isDragActive, isDragReject) {
+    if (!children) {
```
okonet: Why do you need this check?

aymericbouzy: React.cloneElement doesn't like null argument

src/index.js
aymericbouzy: so if you provide a function, the props will not be added to the children. In case some people start mixing together the two syntax, it might seem weird to them ... But you can decide that you should chose between the two syntaxes.

src/index.js
aymericbouzy: additionalProps ?

src/index.js
aymericbouzy: That's what I'm suggesting indeed. Feel free to tell me if you disagree :)

okonet: I actually tend to agree on this. Let's stick to function for now and remove `React.cloneElement`. We'll need an updated README.md that will make this clear.

aymericbouzy: Ok I'm removing it then and refactoring. I'll also suggest an update on the README.md.

aymericbouzy:
Please tell me if you're requesting more changes 🙂

src/index.spec.js
```diff
@@ -281,6 +296,27 @@ describe('Dropzone', () => {
...
+            }
+            return <div id="blank">some content</div>;
+          }}
+        </Dropzone>
+      );
+      expect(dropzone.find('#blank')).not.toBeUndefined();
+      dropzone.simulate('dragEnter', { dataTransfer: { files: images } });
+      expect(dropzone.find('#active')).not.toBeUndefined();
+      dropzone.simulate('dragEnter', { dataTransfer: { files } });
+      expect(dropzone.find('#rejected')).not.toBeUndefined();
```
okonet: I think it would be interesting to also test that `#active` and `#blank` is removed in this case. So for each case we would like to ensure only one of these 3 possible divs is rendered.

README.md
```diff
@@ -68,6 +66,33 @@ var DropzoneDemo = React.createClass({
 React.render(<DropzoneDemo />, document.body);
 ```
 
+Style
+=====
+
+By default, the component picks up some default styling to get you started. You can customize `<Dropzone>` by specifying a `style` and `activeStyle` which is applied when a file is dragged over the zone. You can also specify `className` and `activeClassName` if you would rather style using CSS.
+
+You have a third option : providing a function that returns the component's children.
```
okonet: Not sure if name "Style" fits well. Let's call it "Reacting to user input" or propose a better name.

README.md
```diff
@@ -68,6 +66,33 @@ var DropzoneDemo = React.createClass({
...
+
+```
+<Dropzone>
+  {({ isDragActive, isDragReject }) => {
+    let color = "black", text = "Try dropping some files";
```
okonet: Let's simplify the example. Please remove `color` from it and let's `return` some text in each of the cases just to illustrate it.

okonet:
@aymericbouzy amazing work! It made me think of something I'd like to start a discussion here for.

I'm wondering if we should drop other ways of styling whatsoever and let users decide up on this. The problem right now is that it's impossible to style dropzone with all available methods like CSS-in-JS. For instance I don't think it's possible to style it with https://github.com/styled-components/styled-components ATM. 

This means a breaking change but on the other hand, it will provide much cleaner API IMO. Just FYI I'm planning to drop default styles as of #221 in the near future either way which will also be a breaking change. I believe this PR could be a good transition to a better architecture.

Thoughts?

aymericbouzy:
I was also quite uncomfortable with having to do this : 
```jsx
<Dropzone style={{}} onDrop={this.onDrop}>
  Drop a file
</Dropzone>
```
It's simply weird ...
I think you should have a way for users to opt in for the default style in an easy way, such as : 
```jsx
<Dropzone default_style onDrop={this.onDrop}>
  Drop a file
</Dropzone>
```
But having a component that only handles the "dropping" part, and leaves the "styling" part to the user totally makes sense to me.

aymericbouzy:
Thanks for sharing CSS-in-JS ! I didn't know about it, and it looks really amazing 😃

aymericbouzy:
You may review my code.

okonet:
@aymericbouzy amazing work! The next version will be released automatically soon.

aymericbouzy:
Thanks ! it's not much :) Feel free to ask me for other features if you need. I'll tell you if I can do it.

aymericbouzy:
@okonet Related to CSS-in-JS : have you seen react-css-components ? https://github.com/andreypopp/react-css-components It looks very promising too ! syntax looks clearer to me.

okonet:
Yeah I saw it. Don't like having expressions in CSS though.

Issue #339, opened 2017-01-31
Introduce Flow types


okonet:
Use Flow types Instead of propTypes and export them for better completion support.

okonet:
@ryyppy do you want to help me this one?

pcgilday:
@okonet - Is anyone already working on this? I'm happy to do this if help is still wanted.

okonet:
#344 might be outdated but I'm still interested in merging it so after #468 is done, I'd love some help on this.

SpainTrain:
Have you considered adding defs to `flow-typed` rather than exporting (<https://github.com/flowtype/flow-typed/tree/master/definitions/npm>)?  The main benefit is that `flow-typed` definitions are scoped to `flow` versions.  This allows a library consumer to move to a new `flow` version before its library(ies) do and vice versa.  Otherwise, as an app developer, it is quite possible to get stuck where you either cannot upgrade your libraries or cannot upgrade `flow` due to version conflicts.

It is also easy to get started with `flow-typed` definitions because `flow-typed` will stub out libraries when installed.  If this option sounds interesting, we could start to flesh out our stubs and open PRs on <https://github.com/flowtype/flow-typed/tree/master/definitions/npm>

okonet:
@SpainTrain what's the process for that? Can you help us with that?

ekropotin:
> There is a definition for react-dropzone in flow-typed now: https://github.com/flowtype/flow-typed/blob/master/definitions/npm/react-dropzone_v4.x.x/flow_v0.53.x-/react-dropzone_v4.x.x.js.

There are types for version 4.x while the recent version on NPM is 7.0.1.

rolandjitsu:
@TLadd I understand. I wonder if it would be difficult to have both TS and Flow types in the same repo ... Do you have any idea if this is possible at all?

TLadd:
Sure, it would have to be this repo though (the ts definition is already here: https://github.com/react-dropzone/react-dropzone/blob/master/typings/react-dropzone.d.ts). The other option is to update flow-typed. Just depends on if the react-dropzone maintainers want to have them here or not. Obviously makes it easier to ensure the types are kept up to date, but adds a maintenance cost.

rolandjitsu:
@TLadd I am a maintainer 😄 But I plan on moving the current code to TS, which means the TS typings that currently are separate will be gone. So I'm not sure how the flow types would work with the new TS code 😕

okonet:
[...]

PR #344, opened 2017-02-03
Introduce Flow types.

ming-soon:
Refs #339


src/getDataTransferItems.js
```diff
@@ -1,4 +1,5 @@
-export default function getDataTransferFiles(event, isMultipleAllowed = true) {
+/* @flow */
+export default function getDataTransferFiles(event: Object, isMultipleAllowed: boolean = true) {
```
ryyppy: What exactly is an `event` ? apparently you want to express an object that has a `event.dataTransfer` attribute and a `event.target` attribute... why not make this explicit?

src/index.js
```diff
@@ -1,23 +1,84 @@
...
+  type: string,
+  style: Object,
+  multiple: boolean,
+  name?: string,
+  ref: Function,
+  onChange: Function
+};
+
+export type Props = {
+  onClick: Function,
```
ryyppy: what kind of Function is this? Make this explicit!
E.g. `onClick: (elementId: string) => void`

src/index.js
```diff
@@ -1,23 +1,84 @@
...
     super(props, context);
-    this.onClick = this.onClick.bind(this);
-    this.onDragStart = this.onDragStart.bind(this);
-    this.onDragEnter = this.onDragEnter.bind(this);
-    this.onDragLeave = this.onDragLeave.bind(this);
-    this.onDragOver = this.onDragOver.bind(this);
-    this.onDrop = this.onDrop.bind(this);
-    this.onFileDialogCancel = this.onFileDialogCancel.bind(this);
-    this.fileAccepted = this.fileAccepted.bind(this);
+    (this:any).onClick = this.onClick.bind(this);
```
ryyppy: you need to define `onClick` as a property type in the class body, to make this flow understand the `this.onClick`...

src/index.js
```diff
@@ -27,21 +88,25 @@ class Dropzone extends React.Component {
...
   componentWillUnmount() {
     // Can be replaced with removeEventListener, if addEventListener works
-    document.body.onfocus = null;
+    if (document.body instanceof HTMLElement) {
+      document.body.onfocus = null;
+    }
   }
 
-  onDragStart(e) {
+  onDragStart(e: Object) {
```
ryyppy: I guess `e` is a DOM element... you can leverage the internal flowtype libdefs, which define exactly this: `Element` as defined here: https://github.com/facebook/flow/blob/master/lib/dom.js#L939

okonet: `e` is `event`. Let's rename arguments to make them less ambiguous.

okonet:
@ming-soon thanks for the PR! This is looking promising to me. Let's address feedback from @ryyppy and we're almost there I think.

One more concern is deprecating of some props. We should probably wait with this PR until next major release since I'd like to use https://github.com/Aweary/react-is-deprecated and I don't know how this will work with Flow. Ideas?

src/index.js
ryyppy: okay, so it will be `onDragStart(e: Event)` then :-)

ming-soon: We can use `SyntheticDragEvent` to be more specific.

[https://github.com/facebook/flow/blob/v0.35.0/lib/react.js#L349](https://github.com/facebook/flow/blob/v0.35.0/lib/react.js#L349)

okonet: Yes, let's reuse React's type defs as much as possible.

.babelrc
```diff
@@ -1,4 +1,4 @@
 {
   "presets": ["es2015", "react", "stage-0"],
-  "plugins": ["add-module-exports"]
+  "plugins": ["add-module-exports", "transform-flow-strip-types"]
```
okonet: I'm wondering how we should deliver flow types to npm so people can get advantage of them in their projects? @ryyppy I know you did this before.

src/getDataTransferItems.js
okonet: Let's add return types everythere for such cases: `function(arg: type): type {}`

src/index.js
```diff
@@ -1,23 +1,84 @@
...
+
 const supportMultiple = (typeof document !== 'undefined' && document && document.createElement) ?
   'multiple' in document.createElement('input') :
   true;
 
 class Dropzone extends React.Component {
-  constructor(props, context) {
+  props: Props;
+
+  static defaultProps: {
```
okonet: Can we somehow co-locate this with https://github.com/okonet/react-dropzone/pull/344/files#diff-1fdf421c05c1140f6d71444ea2b27638R367?

src/index.js
```diff
@@ -1,23 +1,84 @@
...
+    maxSize: number,
+    minSize: number
+  };
+
+  state: {
+    isDragActive: boolean,
+    isDragReject?: boolean
+  };
+
+  fileInputEl: Object;
```
okonet: This should be a React node, isn't it?

src/index.js
okonet: Yes, please!

src/index.js
```diff
@@ -27,21 +88,25 @@ class Dropzone extends React.Component {
...
 
-  onDragStart(e) {
+  onDragStart(e: Object) {
     if (this.props.onDragStart) {
       this.props.onDragStart.call(this, e);
     }
   }
 
-  onDragEnter(e) {
+  onDragEnter(e: Object) {
```
okonet: `onDragEnter(e: Object): void`?

src/index.js
```diff
@@ -166,15 +231,15 @@ class Dropzone extends React.Component {
...
     return accepts(file, this.props.accept);
   }
 
-  fileMatchSize(file) {
+  fileMatchSize(file: Object) {
     return file.size <= this.props.maxSize && file.size >= this.props.minSize;
   }
 
-  allFilesAccepted(files) {
+  allFilesAccepted(files: Array<Object>) {
```
okonet: Let's define the `DataTransferItemType` based on https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem and reuse it. Also this method return type is missing (`:boolean`)

.babelrc
ryyppy: hm,... depends on your dependencies... if there are no dependencies, it's easier to consume... otherwise we should publish separate libdefs on flow-typed for now

src/index.js
ming-soon: @okonet I think we should use the `File` type based on [https://developer.mozilla.org/en-US/docs/Web/API/File](https://developer.mozilla.org/en-US/docs/Web/API/File)

src/index.js
ming-soon: This can be `HTMLInputElement`.

.babelrc
ming-soon: @okonet By default, the preset [react](https://babeljs.io/docs/plugins/preset-react/) includes the `transform-flow-strip-types` plugin. And right now it is not possible to disable a specific plugin from preset, AFAIK.
[http://stackoverflow.com/questions/33707189/is-there-an-option-to-disable-a-plugin-in-babel-when-using-a-preset](http://stackoverflow.com/questions/33707189/is-there-an-option-to-disable-a-plugin-in-babel-when-using-a-preset)

src/index.js
```diff
@@ -372,4 +384,31 @@ Dropzone.defaultProps = {
   minSize: 0
 };
 
+Dropzone.propTypes = {
```
okonet: Hmm, the whole point of having Flow types is to remove PropTypes. I'm wondering why did you add them back?

okonet: Hmm, can't see this change for the whole PR but only for the last commit. So, I guess, nevermind my comment.

SaraVieira:
@okonet why hasn't this been merged?

okonet:
I wanted to remove style management before proceeding with this one. But it's taking too much time so I'd appreciate any help

SaraVieira:
In this or style management?

okonet:
@SaraVieira In both I guess since after removing style management this one will need to be updated.

src/index.js
```diff
@@ -316,22 +393,19 @@ Dropzone.propTypes = {
...
   rejectClassName: React.PropTypes.string, // className for rejected state
-
   disablePreview: React.PropTypes.bool, // Enable/disable preview generation
   disableClick: React.PropTypes.bool, // Disallow clicking on the dropzone container to open file dialog
   onFileDialogCancel: React.PropTypes.func, // Provide a callback on clicking the cancel button of the file dialog
-
   inputProps: React.PropTypes.object, // Pass additional attributes to the <input type="file"/> tag
   multiple: React.PropTypes.bool, // Allow dropping multiple files
-  accept: React.PropTypes.string, // Allow specific types of files. See https://github.com/okonet/attr-accept for more information
+  accept: React.PropTypes.string, // Allow specific types of files. See https://github.com/okonet/attrccept for more information
```
b0gok: @okonet `https://github.com/okonet/attrccept` is invalid link and `https://github.com/okonet/attr-accept` is valid. Is it correct change?

okonet: No, this is incorrect. Thanks for mentioning.

rolandjitsu:
Is anyone still gonna work on this? And if so, how will this play with TypeScript? Because I plan on migrating the current codebase to TS after https://github.com/react-dropzone/react-dropzone/pull/717.

capaj:
@rolandjitsu you can't have both flow and typescript in the same project. 
I'd say close this PR and move on with typescript. TS has become the standard for typing in JS community. Even projects like yarn or jest which were the biggest pushers of flow migrated to TS.

rolandjitsu:
@okonet what do you think, should we close this PR?

rxmarbles:
@rolandjitsu I think it's not worth the effort and agree it should be closed. However if we do want to some form of typing we could look to switch to typscript

rolandjitsu:
@rxmarbles ok. We already have TS typings, but I also have started working on migrating the current codebase to TS.

rxmarbles:
Ah yes totally forgot about that 😂. Sweet. Look forward to reviewing that PR.

PR #313, opened 2016-12-23
Add styleguide

okonet:
WIP!

This PR adds the styleguide along with some examples.

Closes #105

PR #364, opened 2017-02-23
Feat: Add deprecation warnings for some props

okonet:
In the next major release, some props will be deprecated. In order to smoothen the migration process deprecation warnings were added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,8 @@
     "react": "^0.14.0 || ^15.0.0"
   },
   "dependencies": {
-    "attr-accept": "^1.0.3"
+    "attr-accept": "^1.0.3",
+    "react-is-deprecated": "^0.1.2"
   },
   "devDependencies": {
     "babel-cli": "^6.9.0",
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,7 @@
 /* eslint prefer-template: 0 */
 import React from 'react';
 import accepts from 'attr-accept';
+import { deprecate } from 'react-is-deprecated';
 import getDataTransferItems from './getDataTransferItems';
 
 const supportMultiple = (typeof document !== 'undefined' && document && document.createElement) ?
@@ -325,16 +326,47 @@ Dropzone.propTypes = {
   onDragOver: React.PropTypes.func,
   onDragLeave: React.PropTypes.func,
 
+  // Contents of the dropzone
   children: React.PropTypes.oneOfType([
     React.PropTypes.node,
     React.PropTypes.func
-  ]), // Contents of the dropzone
-  style: React.PropTypes.object, // CSS styles to apply
-  activeStyle: React.PropTypes.object, // CSS styles to apply when drop will be accepted
-  rejectStyle: React.PropTypes.object, // CSS styles to apply when drop will be rejected
-  className: React.PropTypes.string, // Optional className
-  activeClassName: React.PropTypes.string, // className for accepted state
-  rejectClassName: React.PropTypes.string, // className for rejected state
+  ]),
+
+  // CSS styles to apply
+  style: deprecate(
+    React.PropTypes.object,
+    'Prop style is deprecated. Use function as children to style dropzone and its contents.'
+  ),
+
+  // CSS styles to apply when drop will be accepted
+  activeStyle: deprecate(
+    React.PropTypes.object,
+    'Prop activeStyle is deprecated. Use function as children to style dropzone and its contents.'
+  ),
+
+  // CSS styles to apply when drop will be rejected
+  rejectStyle: deprecate(
+    React.PropTypes.object,
+    'Prop rejectStyle is deprecated. Use function as children to style dropzone and its contents.'
+  ),
+
+  // Optional className
+  className: deprecate(
+    React.PropTypes.string,
+    'Prop className is deprecated. Use function as children to style dropzone and its contents.'
+  ),
+
+  // className for accepted state
+  activeClassName: deprecate(
+    React.PropTypes.string,
+    'Prop activeClassName is deprecated. Use function as children to style dropzone and its contents.'
+  ),
+
+  // className for rejected state
+  rejectClassName: deprecate(
+    React.PropTypes.string,
+    'Prop rejectClassName is deprecated. Use function as children to style dropzone and its contents.'
+  ),
 
   disablePreview: React.PropTypes.bool, // Enable/disable preview generation
   disableClick: React.PropTypes.bool, // Disallow clicking on the dropzone container to open file dialog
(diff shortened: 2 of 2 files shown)


cephirothdy2j:
Heeding the deprecation warnings, this morning I went into my project and changed my implementation of the dropzone component. However, given the absence of style, activeStyle and rejectStyle props, the component falls back to an inline-styled box. If I want to roll my own custom dropzone UI, how should I be handling this?

Before (3.10.0) and after (3.11.0 re-implemented) screenshots attached.

dropzone 3 10 0
dropzone 3 11 0

okonet:
I'm working on the new documentation that will reflect those changes. Default styles will be gone as well, so you will have to implement it either way. For now you can pass `style={{}}` to Dropzone as a workaround.

okonet:
See "Reacting to user input" in the readme as well.

cephirothdy2j:
Thanks. I guess I'll stick with 3.10.0 until those are ready.

P.S. love the plugin! Thanks for maintaining it.

danyim:
I'm also seeing these deprecation warnings after upgrading without a clear alternative. How can you use a function as children to override the default styling?

     Prop className is deprecated. Use function as children to style dropzone and its contents.

okonet:
See ↑

ankitmandhani:
Going forward will there be a way to style the outermost `div` created by Dropzone  component?
Forexample, if I want to pass `height: 100%` to the enclosing `div` how can I do that? I followed the advice in "Reacting to user input" but I still need to pass the `style={{height: '100%'}}` to Dropzone.

okonet:
I was thinking about a `component` prop there you can pass any react component. In this case it will integrate well with any styling method I think.

kachkaev:
Hi @okonet,

Just curious: is there any way to disable deprecation warnings in `>3.11`? I did replace the child component with a function in `3.11`, removed `className` and `activeClassname`, so my component started looking like this:

```js
<Dropzone onDrop={this.onDrop.bind(this)} multiple={false}>
{({ isDragActive }) => (/*...*/)}
</Dropzone>
```

Despite that, the console showed me the following:

```
Prop maxSize is deprecated and will be removed in the next major release
Prop minSize is deprecated and will be removed in the next major release
```

Adding `style={{}}` as you suggested above, created another warning:
```
Prop style is deprecated. Use function as children to style dropzone and its contents.
```

I understand that that the new version is WIP, just raising this because the warnings are a bit distracting. It feels tempting to get rid of them them, but there seems to be no way to do so. I spent over an hour trying to investigate where they were coming from, and finally found this discussion.

If `Dropzone` is a dependency of a dependency, investigation may take even longer. An example of such case is https://github.com/OlegIlyenko/graphiql-workspace/issues/5#issuecomment-283588414.

okonet:
Well, the warnings will go away as you migrate to the next version. I released warnings a bit earlier to make some time for people to start the migration but you're right — it might be hard to see where it is coming from. I'm open to any suggestions.

oshalygin:
Really awesome project @okonet , thanks for all the hard work on it.

A few suggestions:
* The new release should have a major version bump(semver) if there are props that are deprecated.
* Errors should be thrown for props that people are using that are deprecated, giving them an easy way to identify whats going on.  
* Those of us that haven't upgraded to the latest version(IE 3.X.X), shouldn't see the warnings.

The consumers of this component should be referencing the `CHANGELOG.md` which should indicate the deprecated props and give them some indication of what they _should_ be using instead.

Just a few suggestions, totally open to helping as well, but I rather know whats accepted before submitting a PR :)

Thanks again @okonet

okonet:
Thanks!

This project uses semantic release thus there is no change log. Instead it uses https://github.com/okonet/react-dropzone/releases

As for major version: I'll make a bump then I'll remove the deprecated props. Right now you still can use props so there is no breaking change.

shwin7:
Hi @okonet 

If maxSize and minSize are deprecated, how will size validation be done? Thanks.

okonet:
@shwin7 you'll have to do it manually in the `onDrop` callback. Since the behavior of the validation outcome isn't clear and heavily depends on the project's needs, this is the proper way of doing it. So, basically, you'll end up doing something like:

```js
onDrop = (files) => {
  if (files.some(file => file.size >= MAX_SIZE)) {
    this.setState({
      error: `Some files exceed the ${ MAX_SIZE } limit.`
    })
  }
}
```

shwin7:
Sure, that makes sense. Thanks for your response!

leyyinad:
So there's no way to ged rid of the `maxSize`/`minSize` warnings right now, because if I don't set them, their defaults will be automatically applied, right?

If I remove the now deprecated `className` prop, The component will add hardcoded styles which I won't be able to remove. I strongly vote for only adding deprecation warnings for props that actually can be unspecified.

vict-shevchenko:
Hi! I really enjoyed existing declarative way of working with Dropzone. Are you planning to keep it?
What problems solve(benefits give) switching to styling as function as a children?

okonet:
@vict-shevchenko it will become even more declarative this way! Without backing features into the library you can create crazy applications. I don't want to explain all the implications here but trust me this is the right way to evolve :)

rijk:
@okonet is [the documentation](https://github.com/okonet/react-dropzone) up to date with the 'new' way of doing things?

okonet:
@rijk [not yet](https://github.com/okonet/react-dropzone/pull/313). You can still use those props though.

killroy42:
Deprecation warnings aren't really useful if you cannot actually fix the behaviour. I only just started using this component and never used any of the deprecated features, and still get warnings about things I do not control. Normally that only happens if there is a bug in the component itself relative to third party library deprecations. Only mentioning it as I also spend several hours until I found this and if you multiply this by all the people not commenting on it, that's a lot of lost work, likely more then without that warning.

okonet:
@killroy42 I'm trying to make the migration path easier in the future and I'm sorry it didn't work well for you right now. But you're welcome to help maintain this repo and make right decisions at the right time. I'm doing this in my free time and don't have an ability to predict all implications. 

Do you have any constructive feedback on how solve this better?

okonet:
BTW the less helpful thing you can do is to come here and say how many hours of your life I've wasted by doing this free work.

killroy42:
Well, as I mentioned, avoid unavoidable deprecation warnings. Something is only really deprecated if there is an alternative to the functionality, or if the functionality is removed. In this case, it appears there is an alternative but not 100%, such that the warnings become unavoidable. Using the "new syntax/api" should simply not cause the warnings. I know you added warnings to help, but in this case you may have been better of without the warnings. After all, you spend time adding them as well. So it's not just affecting other users, but yourself as well. I'm not posting here to give you a hard time. I'm posting here to give you awareness that your good intentions with the thorough warnings may be misguided. I ONLY posted, since somebody else posted the same issue, and often times maintainers will not act unless there is a critical mass response. As I start using this component, I will happily contribute if it serves my use case.

Thanks for your efforts and time, and don't take github comments personal. That way lies madness!

rijk:
Yeah, don't take it personal, @okonet . Obviously we really like your work, otherwise we wouldn't be here using it!

Issue #147, opened 2016-03-08
Incorrect orientation on image taken with mobile device.


c0cky:
I'm taking a picture with my iPhone 5s and trying to upload a file on my desktop, the images orientation is incorrect.

okonet:
I'm not sure this is related to the component. The only place that might be related to this is here: https://github.com/okonet/react-dropzone/blob/master/src/index.js#L92 and this is also a browser feature.

I've found this StackOverflow article: http://stackoverflow.com/questions/29752779/how-can-i-preview-image-before-uploading-with-the-correct-orientation-with-javas

I'd accept a PR that fixes it, but I'd rather keep the surface of this library minimal.

c0cky:
I was just curious because Dropzone.js added a fix for this on the client side about 29 days ago. No worries, I understand if we want to keep this as minimalist as possible.

For anyone who finds this in the future what I'm going to do is just use [https://github.com/blueimp/JavaScript-Load-Image](url) in the onDrop and it automagically reorients the image based on the exif.

okonet:
:+1:

pukapukan:
If you are storing file in S3, you can use S3 item creation event to trigger Lambda to auto orient image using gm.

Andriy-Kulak:
@camrongodbout can you explain how you ended up the JavaScript-Load-Image package in React? if I follow the package instructions and add the package as a script tag in index.html file, then I cannot use loadImage because it is undefined in my specific component. Also, I cannot import the package in my Component properly. Any quick tips would be greatly appreciated!

manoj-makkuboy:
@Andriy-Kulak I created a wrapper for that purpose.

pavelpashkovsky:
@Andriy-Kulak you can use this repo: (https://github.com/blueimp/JavaScript-Load-Image)

`npm i blueimp-load-image`

then:

`import loadImage from 'blueimp-load-image';`

lounsbrough:
@manoj-makkuboy - `fix-image-rotation` doesn't work in IE11.

adityay-perpetualny:
> I was just curious because Dropzone.js added a fix for this on the client side about 29 days ago. No worries, I understand if we want to keep this as minimalist as possible.
> 
> For anyone who finds this in the future what I'm going to do is just use [https://github.com/blueimp/JavaScript-Load-Image](url) in the onDrop and it automagically reorients the image based on the exif.

Could you please help me in implementing this?

[...]

PR #366, opened 2017-02-28
Reset orientation onDrop

samlogan:
Prevents images rotating incorrectly on drop

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
   },
   "dependencies": {
     "attr-accept": "^1.0.3",
+    "exif-js": "^2.1.1",
     "react-is-deprecated": "^0.1.2"
   },
   "devDependencies": {
--- a/src/index.js
+++ b/src/index.js
@@ -2,6 +2,7 @@
 import React from 'react';
 import accepts from 'attr-accept';
 import { deprecate } from 'react-is-deprecated';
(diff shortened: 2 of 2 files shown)


okonet:
What is this suppose to fix? And why should this be Dropzone's concern? IMO this should live in the userland (ImagePreview component).

samlogan:
I hear you and get your motivation to keep things minimal. However, I am using Dropzone on a recent project and images that were uploaded by a mobile device were rotated incorrectly in the preview, after checking the issue log I found [this](https://github.com/okonet/react-dropzone/issues/147) where you mentioned you would accept a PR that fixes it. Feel free to close this if you feel it is not Dropzone's concern. IMO I would like this to work out of the box.

okonet:
My concern is that it increases the lib size rather significantly for a tiny edge case. https://github.com/blueimp/JavaScript-Load-Image has this solved and probably should be used instead.

okonet:
Another idea could be to implement the image preview component separately and include it as a part of the library.

samlogan:
I like the sound of making the preview a separate component. I can create a new PR with this added

okonet:
Let's try this out and see how it feels.

samlogan:
No stress, will let you know how it goes

Issue #370, opened 2017-03-07
When dragging on child content that is replaced, dragLeave is not called


mwolson:
When the child content can vary according to whether the drag is active, such as putting an overlay in front of it or replacing some content, it's possible to get an initial dragEnter event when you drag onto that content, but you'll never get a corresponding dragLeave event, so the enter counter gets thrown off and the dragActive styles never go away.

A better solution is to track which elements have provided dragEnter events, and then on receiving a dragLeave event, remove from that list any DOM elements which are no longer part of the current container.

Here's a Jest test case to demonstrate the problem. PR will be coming soon.

```js
    it('should reset the dragActive/dragReject state when leaving after a child goes away', () => {
      const DragActiveComponent = () => <p>Active</p>;
      const ChildComponent = () => <p>Child component content</p>;
      const dropzone = mount(
        <Dropzone>
          {({ isDragActive, isDragReject }) => {
            if (isDragReject) {
              return 'Rejected';
            }
            if (isDragActive) {
              return <DragActiveComponent />;
            }
            return <ChildComponent />;
          }}
        </Dropzone>
      );
      dropzone.find(ChildComponent).simulate('dragEnter', { dataTransfer: { files } });
      dropzone.simulate('dragEnter', { dataTransfer: { files } });
      expect(dropzone.state().isDragActive).toEqual(true);
      expect(dropzone.state().isDragReject).toEqual(false);

      dropzone.simulate('dragLeave', { dataTransfer: { files } });
      expect(dropzone.state().isDragActive).toEqual(false);
      expect(dropzone.state().isDragReject).toEqual(false);
    });
  });
```

CC: @tonymet

PR #371, opened 2017-03-07
Fix dragLeave handling when children change (#370)

mwolson:
When the child content can vary according to whether the drag is active, such as putting an overlay in front of it or replacing some content, it's possible to get an initial dragEnter event when you drag onto that content, but you'll never get a corresponding dragLeave event, so the enter counter gets thrown off and the dragActive styles never go away.

* Track which elements have been the targets of `dragEnter` events
* On receiving a `dragLeave` event, remove from that list any DOM elements which are no longer part of the current container
* Add test case for this

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -26,6 +26,7 @@ class Dropzone extends React.Component {
     this.onDrop = this.onDrop.bind(this);
     this.onFileDialogCancel = this.onFileDialogCancel.bind(this);
     this.fileAccepted = this.fileAccepted.bind(this);
+    this.setRef = this.setRef.bind(this);
     this.isFileDialogActive = false;
     this.state = {
       isDragActive: false
@@ -33,7 +34,7 @@ class Dropzone extends React.Component {
   }
 
   componentDidMount() {
-    this.enterCounter = 0;
+    this.dragTargets = [];
     // Tried implementing addEventListener, but didn't work out
     document.body.onfocus = this.onFileDialogCancel;
   }
@@ -53,7 +54,9 @@ class Dropzone extends React.Component {
     e.preventDefault();
 
     // Count the dropzone and any children that are entered.
-    ++this.enterCounter;
+    if (this.dragTargets.indexOf(e.target) === -1) {
+      this.dragTargets.push(e.target);
+    }
 
     const allFilesAccepted = this.allFilesAccepted(getDataTransferItems(e, this.props.multiple));
 
@@ -85,8 +88,9 @@ class Dropzone extends React.Component {
   onDragLeave(e) {
     e.preventDefault();
 
-    // Only deactivate once the dropzone and all children was left.
-    if (--this.enterCounter > 0) {
+    // Only deactivate once the dropzone and all children have been left.
+    this.dragTargets = this.dragTargets.filter(el => el !== e.target && this.node.contains(el));
+    if (this.dragTargets.length > 0) {
       return;
     }
 
@@ -110,7 +114,7 @@ class Dropzone extends React.Component {
     e.preventDefault();
 
     // Reset the counter along with the drag on a drop.
-    this.enterCounter = 0;
+    this.dragTargets = [];
     this.isFileDialogActive = false;
 
     fileList.forEach((file) => {
@@ -172,6 +176,10 @@ class Dropzone extends React.Component {
         }
       }, 300);
     }
+  }
+
+  setRef(ref) {
+    this.node = ref;
   }
 
   fileAccepted(file) {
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -53,7 +53,8 @@ class Dropzone extends React.Component {
     e.preventDefault();
 
     // Count the dropzone and any children that are entered.
-    ++this.enterCounter;
+    this.dragTargets = this.dragTargets.filter(el => el !== e.target);
+    this.dragTargets.push(e.target);
```
okonet: I'm confused by this part TBH. You're filtering something and then you push to the array and mutate it. Can this be a single operation instead? `reduce` for example?

src/index.js
```diff
@@ -86,7 +87,9 @@ class Dropzone extends React.Component {
     e.preventDefault();
 
     // Only deactivate once the dropzone and all children was left.
-    if (--this.enterCounter > 0) {
+    this.dragTargets = this.dragTargets.filter(el => el !== e.target);
+    this.dragTargets = this.dragTargets.filter(el => this.node.contains(el));
```
okonet: Also here. This is very confusing to see 2 filters in a row with results overwriting each other. This can be rewritten like:

```
this.dragTargets =  this.dragTargets.filter(el => el !== e.target && this.node.contains(el));
```

but I'm wondering why do even need to assign it back to this instance variable if you're only concerned about the length of the array?

Wouldn't it easier to use `Array.prototype.some` or `Array.prototype.any` instead and not mutate it?

src/index.js
```diff
@@ -297,6 +300,7 @@ class Dropzone extends React.Component {
         onDragOver={this.onDragOver}
         onDragLeave={this.onDragLeave}
         onDrop={this.onDrop}
+        ref={node => (this.node = node)}
```
okonet: This will generate a new function each time. Instead, define a function outside of the scope and pass a reference to it.

src/index.js
mwolson: I'll change it to be:

```js
if (this.dragTargets.indexOf(e.target) === -1) {
  this.dragTargets.push(e.target);
}
```

src/index.js
mwolson: What you suggested seems fine, will implement:

```js
this.dragTargets =  this.dragTargets.filter(el => el !== e.target && this.node.contains(el));
```

src/index.js
mwolson: OK.

mwolson:
Pushed the requested changes, ready for another round of review.

src/index.spec.js
```diff
@@ -336,6 +336,7 @@ describe('Dropzone', () => {
       );
       dropzone.find(ChildComponent).simulate('dragEnter', { dataTransfer: { files } });
       dropzone.simulate('dragEnter', { dataTransfer: { files } });
+      dropzone.simulate('dragEnter', { dataTransfer: { files } }); // make sure we de-dupe events
```
okonet: Hmm, not sure I get this line. You're entering 2 times and leaving just 1 but asserts are passing? They shouldn't or am I missing something?

mwolson: We want to make sure that even if the browser sends us duplicate `dragEnter` events for the same element, which it might, we'll still handle it correctly without the `dragTargets` list growing too large. It would also be possible to make the `dragEnter` just be a simple push and rely on the filter in `dragLeave` to remove any duplicates later, up to you.

okonet: Makes sense! I think it's fine not to allocate too much memory and save some work later. Let's stick to it.

RaulTsc:
Let me know if you need any help with pushing this forward for a release.

src/index.js
```diff
@@ -106,39 +106,47 @@ class Dropzone extends React.Component {
...
     this.dragTargets = [];
     this.isFileDialogActive = false;
 
-    fileList.forEach((file) => {
-      if (!disablePreview) {
-        file.preview = window.URL.createObjectURL(file); // eslint-disable-line no-param-reassign
-      }
+    // Don't fire callbacks if a child Dropzone component has already handled this event
+    // eslint-disable-next-line no-underscore-dangle
+    if (!e.__dropzoneChildHandledThis) {
```
okonet: Why can't we use `event.stopPropagation` for this?

src/index.js
```diff
@@ -106,39 +106,47 @@ class Dropzone extends React.Component {
...
+      e.persist();
```
okonet: Hmm, what this is doing?

mwolson:
Finished removing the handling of nested dropzone components and squashed the previous work. I added a test case to describe the current behavior, which is that both dropzones will fire onDrop() events.

PR #373, opened 2017-03-08
Disallow dropping file outside of the Dropzone by default (#87)

mwolson:
* New option `allowGlobalDrops` - if true - will allow dropped items to potentially take over the browser window
* If false, the new default behavior, it will instead install document event handlers to prevent this from happening

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ You have a third option : providing a function that returns the component's chil
 Features
 ========
 
+- `preventDropOnDocument` `[Boolean | **true**]` — When a file is dropped outside of any `<Dropzone>` instance, whether to prevent the browser from navigating to it.
 - `disableClick` `[Boolean | **false**]` — Clicking the `<Dropzone>` brings up the browser file picker.
 - `multiple` `[Boolean | **true**]` — Accept multiple files
 - `minSize` `[Number | **0**]` —  Only accept file(s) larger than  `minSize` bytes.
--- a/src/index.js
+++ b/src/index.js
@@ -9,6 +9,11 @@ const supportMultiple = (typeof document !== 'undefined' && document && document
   true;
 
 class Dropzone extends React.Component {
+  static onDocumentDragOver(e) {
+    // allow the entire document to be a drag target
+    e.preventDefault();
+  }
+
   static renderChildren(children, isDragActive, isDragReject) {
     if (typeof children === 'function') {
       return children({ isDragActive, isDragReject });
@@ -19,6 +24,7 @@ class Dropzone extends React.Component {
   constructor(props, context) {
     super(props, context);
     this.onClick = this.onClick.bind(this);
+    this.onDocumentDrop = this.onDocumentDrop.bind(this);
     this.onDragStart = this.onDragStart.bind(this);
     this.onDragEnter = this.onDragEnter.bind(this);
     this.onDragLeave = this.onDragLeave.bind(this);
@@ -34,14 +40,34 @@ class Dropzone extends React.Component {
   }
 
   componentDidMount() {
+    const { preventDropOnDocument } = this.props;
     this.dragTargets = [];
+
+    if (preventDropOnDocument) {
+      document.addEventListener('dragover', Dropzone.onDocumentDragOver, false);
+      document.addEventListener('drop', this.onDocumentDrop, false);
+    }
     // Tried implementing addEventListener, but didn't work out
     document.body.onfocus = this.onFileDialogCancel;
   }
 
   componentWillUnmount() {
+    const { preventDropOnDocument } = this.props;
+    if (preventDropOnDocument) {
+      document.removeEventListener('dragover', Dropzone.onDocumentDragOver);
+      document.removeEventListener('drop', this.onDocumentDrop);
+    }
     // Can be replaced with removeEventListener, if addEventListener works
     document.body.onfocus = null;
+  }
+
+  onDocumentDrop(e) {
+    if (this.node.contains(e.target)) {
+      // if we intercepted an event for our instance, let it propagate down to the instance's onDrop handler
+      return;
+    }
+    e.preventDefault();
+    this.dragTargets = [];
   }
 
   onDragStart(e) {
@@ -283,6 +309,7 @@ class Dropzone extends React.Component {
     // Remove custom properties before passing them to the wrapper div element
     const customProps = [
       'acceptedFiles',
+      'preventDropOnDocument',
       'disablePreview',
       'disableClick',
       'onDropAccepted',
@@ -318,6 +345,7 @@ class Dropzone extends React.Component {
 }
 
 Dropzone.defaultProps = {
+  preventDropOnDocument: true,
   disablePreview: false,
   disableClick: false,
   multiple: true,
@@ -377,6 +405,7 @@ Dropzone.propTypes = {
     'Prop rejectClassName is deprecated. Use function as children to style dropzone and its contents.'
   ),
 
+  preventDropOnDocument: React.PropTypes.bool, // If false, allow dropped items to take over the current browser window
   disablePreview: React.PropTypes.bool, // Enable/disable preview generation
   disableClick: React.PropTypes.bool, // Disallow clicking on the dropzone container to open file dialog
   onFileDialogCancel: React.PropTypes.func, // Provide a callback on clicking the cancel button of the file dialog
--- a/src/index.spec.js
+++ b/src/index.spec.js
@@ -104,6 +104,81 @@ describe('Dropzone', () => {
     });
   });
 
+  describe('document drop protection', () => {
+    let dropzone;
+    let addEventCalls;
+    let savedAddEventListener;
+    let savedRemoveEventListener;
+
+    beforeEach(() => {
+      savedAddEventListener = document.addEventListener;
+      savedRemoveEventListener = document.removeEventListener;
+      document.addEventListener = spy();
+      document.removeEventListener = spy();
+    });
+
+    afterEach(() => {
+      document.addEventListener = savedAddEventListener;
+      document.removeEventListener = savedRemoveEventListener;
+    });
+
+    // Collect the list of addEventListener/removeEventListener spy calls into an object keyed by event name.
+    function collectEventListenerCalls(calls) {
+      return calls.reduce((acc, [eventName, ...rest]) => {
+        acc[eventName] = rest; // eslint-disable-line no-param-reassign
+        return acc;
+      }, {});
+    }
+
+    it('installs hooks to prevent stray drops from taking over the browser window', () => {
+      dropzone = mount(<Dropzone><p>Content</p></Dropzone>);
+      expect(dropzone.html()).toMatchSnapshot();
+      expect(document.addEventListener.callCount).toEqual(2);
+      addEventCalls = collectEventListenerCalls(document.addEventListener.args);
+      Object.keys(addEventCalls).forEach((eventName) => {
+        expect(addEventCalls[eventName][0]).toBeDefined();
+        expect(addEventCalls[eventName][1]).toBe(false);
+      });
+    });
+
+    it('terminates drags and drops on elements outside our dropzone', () => {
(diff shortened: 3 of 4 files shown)


okonet:
Thanks for your PR!

> New option allowGlobalDrops - if true - will allow dropped items to potentially take over the browser window

Not sure how introducing the new prop will solve the #87. Why do you think this can be useful?

mwolson:
In this PR, there are some addEventListener calls for drop and dragover. The new prop controls whether those get installed. When installed (this happens by default, false case) these hooks make the entire browser window an allowable drag-and-drop target, but don't do anything with the drop. This overrides the default browser behavior of navigating to the dropped item.

okonet:
@mwolson great work on this PR! My question is if we even need a prop o control that behavior? I can't imagine when the browser default behavior would be desirable in case there is dropzone installed. Do you?

Second thing: can you add a test where multiple dropzones are being used on the same page and check if it still passes? Did you actually test in the real browser as well? I know this is very cumbersome at the moment.

mwolson:
I'd be OK with removing the prop if you want, since the default browser behavior is definitely not desirable.

I've tested it in a real browser (Firefox and Chrome, not yet IE/Edge) with a single dropzone. I'll add a unit test with multiple dropzones shortly.

mwolson:
One possibility for why we might need a prop in the future: if the user has 1000+ very granular dropzones for some reason, they may want to install the global document handling only once. But I don't know how probable it would be for someone to do that.

mwolson:
After testing, the document drop handler is capturing events meant for Dropzone components, will need to handle targets of components that have gone away during re-render, and parent Dropzone components need special handling to reset their state when child Dropzone components get a drop. Going to work further to address these before proceeding with a merge.

mwolson:
Handled child/parent drops in 4b2d86b on my other PR.

mwolson:
Rebased this on top of my other branch since there's starting to be unavoidable overlap. Finished addressing all of the remaining issues with both PR's. Removed WIP status - ready for another look.

README.md
```diff
@@ -92,6 +92,7 @@ You have a third option : providing a function that returns the component's chil
 Features
 ========
 
+- `allowGlobalDrops` `[Boolean | **false**]` — Dropping files outside of the `<Dropzone>` is allowed. This can cause the current browser window to be taken over by the dropped file.
```
okonet: Should this be renamed to something like `preventDropOnDocument` with the default value of `true`? I think right the purpose of it is not clear enough.

src/index.js
```diff
@@ -19,30 +24,52 @@ class Dropzone extends React.Component {
...
     };
   }
 
   componentDidMount() {
-    this.enterCounter = 0;
+    const { allowGlobalDrops } = this.props;
+    this.dragTargets = [];
+
+    if (!allowGlobalDrops) {
+      document.addEventListener('dragover', Dropzone.onDocumentDragOver, false);
```
okonet: Why do we need the `dragover` handler?

src/index.js
```diff
@@ -102,39 +132,47 @@ class Dropzone extends React.Component {
...
+    this.dragTargets = [];
     this.isFileDialogActive = false;
 
-    fileList.forEach((file) => {
-      if (!disablePreview) {
-        file.preview = window.URL.createObjectURL(file); // eslint-disable-line no-param-reassign
-      }
+    // Don't fire callbacks if a child Dropzone component has already handled this event
+    // eslint-disable-next-line no-underscore-dangle
+    if (!e.__dropzoneChildHandledThis) {
```
okonet: Hmm, I don't like how this looks TBH. Why not use `event.stopPropagation` instead?

src/index.js
```diff
@@ -102,39 +132,47 @@ class Dropzone extends React.Component {
...
+      e.persist();
```
okonet: Can you point me at what this does?

src/index.spec.js
```diff
@@ -37,6 +49,53 @@ describe('Dropzone', () => {
       expect(dropzone.html()).toMatchSnapshot();
     });
 
+    function collectEventListenerCalls(calls) {
```
okonet: Please add a JSDoc comment for this function. I hardly understand what it does and why it's needed.

src/index.spec.js
```diff
@@ -37,6 +49,53 @@ describe('Dropzone', () => {
...
+      return calls.reduce((acc, [eventName, ...rest]) => {
+        acc[eventName] = rest; // eslint-disable-line no-param-reassign
+        return acc;
+      }, {});
+    }
+
+    it('prevents stray drops from taking over the browser window by default', () => {
+      const dropzone = mount(<Dropzone><p>Content</p></Dropzone>);
+      expect(dropzone.html()).toMatchSnapshot();
+      expect(document.addEventListener.callCount).toEqual(2);
```
okonet: It's probably better to test in a separate `it` block with an appropriate description. Either way, shouldn't it test explicitly what handlers were attached?

src/index.spec.js
```diff
@@ -37,6 +49,53 @@ describe('Dropzone', () => {
...
+      const addCalls = collectEventListenerCalls(document.addEventListener.args);
```
okonet: ?

src/index.spec.js
```diff
@@ -37,6 +49,53 @@ describe('Dropzone', () => {
...
+
+      const event = { preventDefault: spy() };
```
okonet: I wondering if this test should be that detailed. I think it will be enough to test that we've called the callback function for appropriate edge cases?

src/index.spec.js
```diff
@@ -317,23 +376,59 @@ describe('Dropzone', () => {
       dropzone.simulate('dragEnter', { dataTransfer: { files } });
       expect(dropzone.text()).toEqual('Rejected');
     });
+
+    it('should reset the dragActive/dragReject state when leaving after a child goes away', () => {
```
okonet: Wasn't it the separate issue?

src/index.spec.js
```diff
@@ -498,6 +593,144 @@ describe('Dropzone', () => {
       expect(dropRejectedSpy.callCount).toEqual(0);
     });
 
+    it('accepts files on the innermost Dropzone when we drop into it', () => {
```
okonet: Not sure what this is testing. What is `innermost` Dropzone?

README.md
mwolson: `preventDropOnDocument` seems fine

src/index.js
mwolson: If `dragover` is not provided, dropped items will still take over the browser window, because the browser doesn't consider a div with only a `drop` event (and no `dragover`) to be a valid `drop` target.

src/index.js
mwolson: The problem is that the parent needs to know whether some child has already fired an onDrop() callback. If a child has fired, reset the dragActive/dragReject/dragTargets state and don't fire onDrop(). If a child has not fired, reset the state and do fire onDrop().

If we used `event.stopPropagation()`, the parent would never receive the event, so its state would remain active, and the render of the parent would still be styled with `dragActive=true` after a drop on the child.

The order is like:
1: Document `drop` handler (if installed via preventGlobalDrops=true)
2: Child `drop` handler
3: Parent `drop` handler

src/index.js
mwolson: Normally event objects that React passes to event handlers are structurally reused, like:

```js
delete e.target;
```

And then put in a pool. If another event comes in, they'll get removed from the pool and sent to another handler.

`e.persist()` will keep that event out of the pool, so we get our own copy of it, and can mutate the event by adding the `__dropzoneChildHandledThis` property. When I tried to do this without an `e.persist()` call, React would throw a warning to the browser console.

Further reading: https://facebook.github.io/react/docs/events.html

src/index.spec.js
mwolson: OK I'll add JSDoc

src/index.js
okonet: Wait a minute, are you talking about nested Dropzones? Why would we support that? What's the use case?

src/index.spec.js
mwolson: I'll split this into its own `describe` block I think, with multiple `it` steps.

src/index.spec.js
mwolson: The problem is that there's no way to fire an event on `document`. So if we didn't manually exercise it here, we wouldn't get test coverage. And it's useful to make sure that we don't call `e.preventDefault()` in the case where the drag event is meant for one of our dropzones - because if we did call `e.preventDefault()` in that case, the dropzone would not get the event.

```
1: document onDrop() <-- stops propagation -->
2: dropzone instance onDrop() <-- doesn't get event now -->
```

src/index.spec.js
mwolson: The innermost dropzone is the child Dropzone component. The outermost dropzone is the parent Dropzone component.

okonet: Yeah, as I already asked ↑ do we really need to support that case? It seems very edgy to me and the complexity it brings to the project doesn't seem worth it. What's the use case for it?

src/index.js
mwolson: Yeah it's nested Dropzones. An example would be: an instant messenger UI where dragging over an entire messaging window will show an overlay modal, and then you may potentially be able to drop into one or more large icons to indicate whether to send the attachment immediately or compose a message to go with it, with a default behavior of "compose" if the user drops into the modal itself and misses the icons.

okonet: I don't believe the complexity that we need to bring in to support such a edge case is worth it TBH. Let's keep things simple as long as possible. When I was asking about multiple dropzones on the page I was referring to a use case with a list of dropzones for instance. Are you okay with that?

src/index.js
okonet: Interesting. I'll read but see my ↑ comment.

src/index.js
mwolson: OK. I don't immediately need nested dropzone support for my own use case yet, but might in the future. I think it might be surprising for someone who tries nesting to find that the parent still receives a drop event, but it's hard to predict whether anyone will actually want to do that. I'll go ahead and remove it from the 2 PR's and stash the changes somewhere; maybe it could be a separate PR by itself later if needed.

I tested the sibling case manually, and it works fine. A single user drop gets fired like:

```
event chain 1: sibling 1 document drop: <-- this isn't me, e.preventDefault() -->
event chain 2: sibling 2 document drop: <-- don't call e.preventDefault() -->
event chain 2: sibling 2 dropzone instance: <-- receives event -->
```

okonet: Awesome!

> I think it might be surprising for someone who tries nesting to find that the parent still receives a drop event

TBH I would be surprised if it wouldn't receive and fire the event :)

mwolson:
Finished one round of changes, still need to split out parent/child nesting, will do later today.

mwolson:
Finished rebasing this on top of the other updated PR, so there should be no more handling of nested components (except for the new tests which document the current behavior).

okonet:
Can you please rebase on top of master?

mwolson:
Sure thing, finished the rebase onto master

okonet:
@mwolson just wondering if you also tries this example:

```js
function dragover_handler(ev) {
  console.log("dragOver");
  // Prevent default select and drag behavior
  ev.preventDefault();
}
```

from https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/File_drag_and_drop

It says it should prevent default browser behavior.

mwolson:
@okonet that example prevents default browser dragover behavior, but only for drags on top of that div. It doesn't do anything for drags on the rest of the document, so we still need the `document` handlers.

okonet:
So even if this handler is on `document`? I'm just wondering if we can go away with just one handler on the document.

tt-mikeolson:
Both handlers are needed - take either away and the feature stops working.

mwolson:
Both handlers are needed - take either away and the feature stops working.

okonet:
Gotcha! Thanks!

Issue #362, opened 2017-02-21
Firefox 51.0.1  activates rejectStyle even on accepted file types.


releaf:
In Firefox 51.0.1, the dropzone initiates the rejectStyle even though a file is of the accepted type. The drop and upload still works fine but the UI indicates otherwise. 

![fire-fox-bug](https://cloud.githubusercontent.com/assets/4107042/23179641/93008344-f83c-11e6-9fe4-9e81034116ca.gif)

okonet:
Looks like Firefox broke it. I believe it should be fixed in Firefox since it looks like they broke the HTML5 API. Fancy to submit a bug to them?

releaf:
Hopefully this gets some traction. Please feel free to add some clarity around the issue if you can. 
https://bugzilla.mozilla.org/show_bug.cgi?id=1342057

mwolson:
None of the workarounds proposed in the Firefox bugtracker are any good, since they happen on a `drop()` event, and not on `dragEnter` events. Until they fix it upstream, we should at least make it look like all attachments are accepted while dragging, which brings us to parity with Safari (which doesn't support file types on drag items at all). A false positive is better than a false negative here. PR #382 implements this.

PR #382, opened 2017-03-14
fix: Don't display bogus Firefox drag types as rejected

mwolson:
Since Firefox always returns a type of 'application/x-moz-file', until https://bugzilla.mozilla.org/show_bug.cgi?id=1342057 is implemented correctly on their side, the best we can do is treat that type as always accepted. A false positive is better than a false negative in this case, and brings behavior parity with Safari, which does not support file types at all.

- [X] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [X] Not relevant

**Summary**

Provide a workaround for https://github.com/okonet/react-dropzone/issues/362 that leads to less user confusion.

No

**Other information**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -206,7 +206,7 @@ class Dropzone extends React.Component {
   }
 
   fileAccepted(file) {
-    return accepts(file, this.props.accept);
+    return file.type === 'application/x-moz-file' || accepts(file, this.props.accept);
   }
 
   fileMatchSize(file) {
--- a/src/index.spec.js
+++ b/src/index.spec.js
@@ -599,6 +599,30 @@ describe('Dropzone', () => {
       expect(dropRejectedSpy.callCount).toEqual(0);
     });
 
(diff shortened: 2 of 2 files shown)


okonet:
Should it go to https://github.com/okonet/attr-accept?

okonet:
Just FYI this behavior is fixed in Firefox 53+

mwolson:
In this case, the workaround is for dragEnter only, while attr-accept is more generic, so react-dropzone would probably be the best place for it. Until Firefox 53+ is actually released, I'd need to keep a forked copy around, so it's up to you.

mwolson:
I pushed c56d2a6 to this branch in my repo to add a comment.

When Firefox 53 comes out, they will start to pass correct MIME types, and that `x-moz-file` type will hopefully never be seen again.

okonet:
Great! Can you create a new PR to merge it to master? Or is it there already?

mwolson:
OK, filed PR #388 for that

PR #392, opened 2017-03-24
fix: make click on dropzone work on safari mobile

martinratinaud:

- [X] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [X] Yes, I've updated the documentation
- [ ] Not relevant

**Summary**

As safari on mobile devices does not handle well the click on an hidden input file, this repository implements the following technique.

`<input type="file" />` fills completely its parent width and height and its opacity is set to 0.
This way, on any mobile device click on dropzone area will work as input is not hidden.

Input container will automatically append a `position: relative` if `position` is not set in the style or className  

Don't think so but every situation has not been tested

**Other information**


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "react-dropzone",
+  "version": "0.0.1",
```
okonet: Please remove this since it's handled by semantic release

README.md
```diff
@@ -151,6 +151,16 @@ var DropzoneDemo = React.createClass({
 React.render(<DropzoneDemo />, document.body);
 ```
 
+Upload on Safari mobile
```
okonet: Why do we need this explanation in the README? I think it's better be placed as a comment in the source code since users should not be concerned about this IMO.

okonet:
I've just tested it and it works as expected with Mobile Safari on iOS 10 so I'm not sure what this suppose to fix.

martinratinaud:
Hi and thanks for your answer.

In my project, click on safari leads to no action. In fact `this.fileInputEl.click();` does not produce anything. My project uses Server Side rendering which may cause this problem.

Feel free to close this PR if nobody gets the same problem as I do and thanks for your time.

README.md
martinratinaud: I put this here because all dropzone component I found get the same problem with safari mobile and I found this important and relevant to tell potential users that this component does work in all cases.

okonet:
SSR shouldn't be an issue. It would be great if you could reproduce this in an isolated env. I'll close this PR for now. Please open an issue if you can reproduce it outside of your project. A GH repo would be great.

martinratinaud:
For the record, I found a solution.

I was using `react-starter-kit` which used `fastclick`. This must have been the problem as when I switched to `react-fastclick` it worked 

See this PR https://github.com/kriasoft/react-starter-kit/pull/1198

Thanks

PR #401, opened 2017-04-04
Only generate a preview after the file has been accepted

ncphillips:

- [X] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [X] Not relevant

**Summary**

I have an orderable list of images inside of a dropzone.

If disablePreview=false and I re-order an image, react-dropzone throws an error because what's beign dragged around is actually a DataTransferItem.

A preview cannot be made from a DataTransferItems, it throws an error. 

> Uncaught TypeError: Failed to execute 'createObjectURL' on 'URL': No function was found that matched the signature provided.

Since the DTI will be rejected by the `this.fileAccepted`, I made it so previews are only made for accepted files.

I have not opened an issue, or found an issue before.

I am not sure. I do not preview rejected files, but perhaps some other people do. If so, suggested at least try-caching that call.

**Other information**

Here's a demo you can do in browser, proving that previews of DataTransferItems cannot be created.

```javascript
let x = { kind: "string", type: "text/html" }
x.__proto__ = DataTransferItem.prototype
window.URL.createObjectURL(x) // Uncaught TypeError: Failed to execute 'createObjectURL' on 'URL': No function was found that matched the signature provided.
```


okonet:
That's awesome! Thanks for contributing.

A few thoughts:

1. Your code will still fail if `accept` prop isn't set: https://github.com/okonet/react-dropzone/blob/master/src/index.js#L210.
1. Which makes me think if we should we just use `try ... catch` block for `createObjectURL` or even better check if the method exists before calling it? I assume this will solve your issue as well?
1. I think this is a breaking change and I think there might be use cases to show previews even if a file was rejected.

ncphillips:
I am not sure if the first point holds. When I modified the source and ran it in my application it seemed to work, even though I don't set `accept`.

That said try-catch might be the right thing to prevent API breaking change

ncphillips:
How would you like to handle this kind of error in the `catch`?

src/index.js
```diff
@@ -138,10 +138,15 @@ class Dropzone extends React.Component {
     this.isFileDialogActive = false;
 
     fileList.forEach((file) => {
-      if (this.fileAccepted(file) && this.fileMatchSize(file)) {
-        if (!disablePreview) {
+      if (!disablePreview) {
+        try {
           file.preview = window.URL.createObjectURL(file); // eslint-disable-line no-param-reassign
+        } catch (err) {
+          // Continue regardless of error
```
okonet: I'm wondering if we should display a warning at least in a dev mode? I think you could re-use `invariant` here and wrap it into `process.env.NODE_ENV !== 'production'` condition. Thoughts?

src/index.js
```diff
@@ -138,10 +138,15 @@ class Dropzone extends React.Component {
     this.isFileDialogActive = false;
 
     fileList.forEach((file) => {
-      if (this.fileAccepted(file) && this.fileMatchSize(file)) {
-        if (!disablePreview) {
+      if (!disablePreview) {
```
okonet: Makes me think if we should tackle the bigger problem in this PR: what if I want to `disablePreviews` based on file property. I was thinking about a callback that can be defined by the user that should return `true` or `false`. This way we could also use it to check if the method exists for this `file` instance and call it.

The problem here I think is that we mutate `file` by adding a new property to the object. But I'm wondering if this is an issue in this context.

Thoughts?

src/index.js
ncphillips: I definitely think that's a good idea. Silent failures are annoying.

src/index.js
ncphillips: A cool idea. 

For my own selfish reasons–we have a feature release Friday that's waiting on this bug fix–I'm not super eager to take on this behaviour at the moment.

It could be part of a follow up PR though!

src/index.js
okonet: Cool, @ncphillips! Will you add it in this PR?

src/index.js
okonet: We have a deal! :)

src/index.js
ncphillips: Yup!

ncphillips: What do you mean by  "re-use invariant"?

I don't see it in the project anywhere.

okonet: Sorry confused with another project where I use it. Yes, add it as a dependency please.

ncphillips: So, using invariant actually breaks the test I added: 'it should not throw an error when failing to build a preview'.

I'm not sure this is the right use-case for invariant.

okonet: Hmm, I think it's okay to throw an error in development mode. Alternatively we could just use `console.error` in this case. What do you think is more appropriate?

ncphillips: I think `console.error` or `console.info` during development would be more appropriate then throwing an error.

With that said, we could provide a `onPreviewFailure` callback so the parent component could handle it however they saw fit. For example, they might want to display an error message to the user.

okonet: Let's address the callback issue in a separate PR since IMO this is user responsibility to not to get to this state. I'd say console.error ahould do. for now.

src/index.js
```diff
@@ -142,7 +142,10 @@ class Dropzone extends React.Component {
         try {
           file.preview = window.URL.createObjectURL(file); // eslint-disable-line no-param-reassign
         } catch (err) {
-          // Continue regardless of error
+          if (process.env.NODE_ENV !== 'production') {
+            const c = console;
```
ncphillips: ES-Lint is cranky.

src/index.js
```diff
@@ -139,7 +139,14 @@ class Dropzone extends React.Component {
 
     fileList.forEach((file) => {
       if (!disablePreview) {
-        file.preview = window.URL.createObjectURL(file); // eslint-disable-line no-param-reassign
+        try {
+          file.preview = window.URL.createObjectURL(file); // eslint-disable-line no-param-reassign
+        } catch (err) {
+          if (process.env.NODE_ENV !== 'production') {
+            const c = console;
+            c.error('Failed to generate preview for file', file, err);
```
okonet: Just disable this line with `// eslint-disable-line no-console`

package.json
```diff
@@ -53,7 +53,8 @@
     "react": "^0.14.0 || ^15.0.0"
   },
   "dependencies": {
-    "attr-accept": "^1.0.3"
+    "attr-accept": "^1.0.3",
+    "invariant": "^2.2.2"
```
okonet: Seems this isn't needed anymore? Please remove it.

ncphillips: woops!

ncphillips:
I accidentally committed a `yarn.lock`. Is that fine?

okonet:
I can live this that :)

Thanks for the PR!

PR #405, opened 2017-04-12
Use `prop-types` instead of `React.PropTypes`

jochenberger:
`React.PropTypes` is deprecated as of React 15.5.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,8 @@
     "react": "^0.14.0 || ^15.0.0"
   },
   "dependencies": {
-    "attr-accept": "^1.0.3"
+    "attr-accept": "^1.0.3",
+    "prop-types": "^15.5.7"
   },
   "devDependencies": {
     "babel-cli": "^6.9.0",
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,6 @@
 /* eslint prefer-template: 0 */
 import React from 'react';
+import PropTypes from 'prop-types';
 import accepts from 'attr-accept';
 import getDataTransferItems from './getDataTransferItems';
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -53,7 +53,8 @@
     "react": "^0.14.0 || ^15.0.0"
   },
   "dependencies": {
-    "attr-accept": "^1.0.3"
+    "attr-accept": "^1.0.3",
+    "prop-types": "^15.5.7"
```
okonet: Should this go to `devDependencies` and `peerDependencies` since it's React's dependency and not ours?

okonet:
Will it work with older React versions or this must be a breaking change?

package.json
jochenberger: See https://www.npmjs.com/package/prop-types#how-to-depend-on-this-package

jochenberger:
See https://www.npmjs.com/package/prop-types#compatibility

andrevenancio:
I just pulled from `npm` version 3.12.3 and on the `src` folder there's still React.PropTypes instead of PropTypes...

okonet:
Probably because of https://github.com/okonet/react-dropzone/blob/master/src/index.js#L378?

@jochenberger mind creating a new PR that removes it as well?

andrevenancio:
Not sure if that's the reason. shouldn't I have the exact same code I see here on the master branch on git when I pull that tgz file from npm? I've pulled it, and the src/index.js still has `React.PropTypes`

jochenberger:
The relevant change is not in 3.12.3 AFAICT.

okonet:
I think `chore` commit type doesn't trigger a release. I'll create an empty commit with `fix` to trigger a build

okonet:
Or rebase master 😱

PR #410, opened 2017-04-19
Show rejected styles when dragging multiple files and multiple==false

### LINKED ISSUE #409: When multiple is set to false the rejectClassName is not shown when multiple files are dragged in.

- [x] I found a bug

1. Set multiple to false and have a rejectClassName
2. Drag 2 files onto the dropzone

The reject class is added to the dropzone element.

plee-nm:
…dragged in multi mode false.

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [ ] Not relevant

**Summary**

Yes!

**Other information**
Closes #409


src/index.js
```diff
@@ -79,11 +79,14 @@ class Dropzone extends React.Component {
       this.dragTargets.push(e.target);
     }
 
-    const allFilesAccepted = this.allFilesAccepted(getDataTransferItems(e, this.props.multiple));
+    // validate all files regardless if they are in multi file mode.
+    const files = getDataTransferItems(e, true);
```
okonet: 1. Shouldn't it be aligned with https://github.com/okonet/react-dropzone/pull/410/files#diff-1fdf421c05c1140f6d71444ea2b27638R133?
1. `true` is default argument, so you can skip it

src/index.js
```diff
@@ -79,11 +79,14 @@ class Dropzone extends React.Component {
...
+    const allFilesAccepted = this.allFilesAccepted(files);
+    const isMultipleNotAllowed = this.props.multiple ? false : files.length > 1;
```
okonet: 1. Let's remove negation from the variable name please
1. This can be rewritten as `!this.props.multiple || files.length > 1`

src/index.js
```diff
@@ -79,11 +79,14 @@ class Dropzone extends React.Component {
...
 
     this.setState({
       isDragActive: allFilesAccepted,
-      isDragReject: !allFilesAccepted
+      isDragReject: !allFilesAccepted || isMultipleNotAllowed
```
okonet: Can `allFilesAccepted` be set accordingly before so this code remains intact?

okonet:
This seems to be a breaking change, right?

plee-nm:
@okonet Yeah it is a breaking change.  What are your thoughts on adding the extra files on drag to the rejected files or all the files as rejected if it is multiple?  Not sure how everyone feels about this but I don't want to have the user experience of a drop of N number of files and it only picking one and saying it is valid.  

We can do this as breaking change or we can add another config like allowMultipleDrop and default it to true.  Setting it to false would show the error class on drag and show any extra files as rejected files.  Or we label it as a breaking change and bump the major version.  Thoughts?  I can submit a PR with these changes.

plee-nm:
@okonet I would push for just making a breaking change, the code will be cleaner and I think the behavior is better for the end user.  Also the breaking change would only affect users using multiple=false and it can be easily documented in a change log.

okonet:
I think it makes sense to make a major version bump. The change you're proposing makes it more logical.

plee-nm:
Alright great I'll submit a pr with these changes.

src/index.js
plee-nm: Not sure what this means exactly.  I just pushed a change so let me know if I need to update it.

src/index.js
plee-nm: I removed the negation with the new changes.

plee-nm:
I just pushed the changes needed to add extra files to rejected when multiple is set to false.  Please let me know what the process is for a major version bump.

src/index.js
```diff
@@ -159,6 +158,11 @@ class Dropzone extends React.Component {
       }
     });
 
+    if (!multiple) {
+      // if not in multi mode add any extra accepted files to rejected.  This will allow end users to easily ignore a multi file drop in "single" mode.
+      rejectedFiles.push(...acceptedFiles.splice(1));
```
okonet: Can this be rewritten with array destructuring?

src/index.js
```diff
@@ -159,6 +158,11 @@ class Dropzone extends React.Component {
       }
     });
 
+    if (!multiple) {
+      // if not in multi mode add any extra accepted files to rejected.  This will allow end users to easily ignore a multi file drop in "single" mode.
```
okonet: Please split the comment into 2 lines

src/index.spec.js
```diff
@@ -455,44 +455,58 @@ describe('Dropzone', () => {
       expect(dropzone.state().isDragReject).toEqual(false);
     });
 
-    it('should expose acceptedFiles and rejectedFiles to children', () => {
+    it('multiple false valid files adds to rejected files on a multple drop', () => {
```
okonet: Can we still keep `should`s in tests?

okonet:
Don't worry about the version bump. I'll take care of this during the merge!

src/index.js
plee-nm: Change acceptedFiles to a `let` so it can be redefined and then do something like 

```
const [accepted, ...rejected] = acceptedFiles;
rejectedFiles.push(...rejected);
acceptedFiles = [accepted];
```

Or just leave it as it is?

okonet: Let's leave as is. Thanks for clarifying.

okonet:
Hey @plee-nm I've just added some more examples. Could you please check you branch out, run `npm start` and test all the examples if they still work? Since you can live-change the examples, try out different states.

plee-nm:
@okonet Just ran through the all the examples, looks like everything is working.

okonet:
I'm looking at how it works now and it doesn't seem to be any breaking changes. We'll still call `onDrop` with the single file. I think it can be released as patch. Am I missing something?

plee-nm:
@okonet it is kind of a gray area because it is fixing undefined behavior.  Making it a major would make sure no one that is relying on the undefined behavior is broken.   Patch or Major I would have some kind of release note saying:  when multiple=false on a multiple file drag it will now show the reject styles/class and add any extra files to the rejected files.

okonet:
Thanks for clarification. I think note is enough since we aren't breaking the API.

PR #427, opened 2017-05-12
Handle changes in the accept setting mid-drag

fritz-c:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Not relevant

**Summary**
When changing the `accept` prop for the component mid-drag, the component did not reflect the change until you happened to enter/leave another element. This PR updates the `isDragActive` state at the timing of the prop change, using `componentWillReceiveProps`, so the state will be updated right away.

In my use case, I was implementing a fullscreen dropzone as described in the documentation, and changed the accept type depending on which region of the page was hovered (which I achieved with a custom `onDragOver` handler). I had set custom classes with `activeClassName`/`rejectClassName`, but they would not reflect the updated `accept` setting until I dragged out and in again.

No

**Other information**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -224,7 +224,7 @@ class Dropzone extends React.Component {
   }
 
   allFilesAccepted(files) {
-    return files.every(file => this.fileAccepted(file));
+    return files.every(this.fileAccepted);
   }
 
   /**
--- a/src/index.js
+++ b/src/index.js
@@ -30,6 +30,7 @@ class Dropzone extends React.Component {
     this.isFileDialogActive = false;
     this.state = {
       isDragActive: false,
+      isDragReject: false,
       acceptedFiles: [],
       rejectedFiles: []
     };
@@ -45,6 +46,19 @@ class Dropzone extends React.Component {
     }
     // Tried implementing addEventListener, but didn't work out
     document.body.onfocus = this.onFileDialogCancel;
+  }
+
+  componentWillReceiveProps(nextProps) {
+    // If the accept setting has changed, and component is currently in the accept or reject states
+    if (nextProps.accept !== this.props.accept && (this.state.isDragActive || this.state.isDragReject)) {
+      const allFilesAccepted = this.allFilesAccepted(this.draggedFiles, nextProps);
+      const isMultipleAllowed = nextProps.multiple || this.draggedFiles.length <= 1;
+
+      this.setState({
+        isDragActive: allFilesAccepted,
+        isDragReject: !allFilesAccepted || !isMultipleAllowed
+      });
+    }
   }
 
   componentWillUnmount() {
@@ -80,9 +94,9 @@ class Dropzone extends React.Component {
       this.dragTargets.push(e.target);
     }
 
-    const files = getDataTransferItems(e);
-    const allFilesAccepted = this.allFilesAccepted(files);
-    const isMultipleAllowed = this.props.multiple || files.length <= 1;
+    this.draggedFiles = getDataTransferItems(e);
+    const allFilesAccepted = this.allFilesAccepted(this.draggedFiles);
+    const isMultipleAllowed = this.props.multiple || this.draggedFiles.length <= 1;
 
     this.setState({
       isDragActive: allFilesAccepted,
@@ -122,6 +136,9 @@ class Dropzone extends React.Component {
       isDragActive: false,
       isDragReject: false
     });
+
+    // Clear files value
+    this.draggedFiles = null;
 
     if (this.props.onDragLeave) {
       this.props.onDragLeave.call(this, e);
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -220,18 +240,18 @@ class Dropzone extends React.Component {
     this.node = ref;
   }
 
-  fileAccepted(file) {
+  fileAccepted(file, props = this.props) {
     // Firefox versions prior to 53 return a bogus MIME type for every file drag, so dragovers with
     // that MIME type will always be accepted
-    return file.type === 'application/x-moz-file' || accepts(file, this.props.accept);
+    return file.type === 'application/x-moz-file' || accepts(file, props.accept);
```
okonet: Let's make `accept` a required second argument and remove defaulting to `this.props` + provide it everywhere explicitly.

src/index.js
```diff
@@ -47,6 +48,19 @@ class Dropzone extends React.Component {
     document.body.onfocus = this.onFileDialogCancel;
   }
 
+  componentWillReceiveProps(nextProps) {
+    // If the accept setting has changed, and component is currently in the accept or reject states
+    if (nextProps.accept !== this.props.accept && (this.state.isDragActive || this.state.isDragReject)) {
+      const allFilesAccepted = this.allFilesAccepted(this.draggedFiles, nextProps);
```
okonet: Can this code duplication be somehow removed? Let's introduce a new method that would handle this for both cases based on arguments.

okonet: By looking at this I think it would make more sense to put dragged file into state and calculate `isDragActive` and `isDragReject` in the `render` method since this is very cheap. Thoughts?

src/index.js
```diff
@@ -220,18 +240,18 @@ class Dropzone extends React.Component {
...
+    return file.type === 'application/x-moz-file' || accepts(file, props.accept);
   }
 
   fileMatchSize(file) {
     return file.size <= this.props.maxSize && file.size >= this.props.minSize;
   }
 
-  allFilesAccepted(files) {
-    return files.every(this.fileAccepted);
+  allFilesAccepted(files, props = this.props) {
```
okonet: Same here, let's make the `accept` a required second arguement to make those functions pure.

src/index.js
fritz-c: I modified the code to do just that. Check it out and see if it fits your expectations.

src/index.js
```diff
@@ -240,18 +213,18 @@ class Dropzone extends React.Component {
     this.node = ref;
   }
 
-  fileAccepted(file, props = this.props) {
+  fileAccepted(file) {
```
okonet: Could this function be a pure one? Let's remove the dependency on `this` here and extract it from the class. There is no reason for it to be a class instance method.

src/index.js
```diff
@@ -292,7 +265,10 @@ class Dropzone extends React.Component {
       ...props // eslint-disable-line prefer-const
     } = rest;
 
-    const { isDragActive, isDragReject } = this.state;
+    const { draggedFiles } = this.state;
+    const isMultipleAllowed = multiple || draggedFiles.length <= 1;
+    const isDragActive = draggedFiles.length !== 0 && this.allFilesAccepted(draggedFiles);
```
okonet: I'd prefer a comparison like `draggedFiles.length > 0` here and on the next line. I think it shows the intention better.

src/index.js
```diff
@@ -292,7 +265,10 @@ class Dropzone extends React.Component {
       ...props // eslint-disable-line prefer-const
     } = rest;
 
-    const { isDragActive, isDragReject } = this.state;
+    const { draggedFiles } = this.state;
+    const isMultipleAllowed = multiple || draggedFiles.length <= 1;
```
okonet: I know it's a nit peak but you're accessing `length` property 3 times in a row. Should it be extracted, you think?

src/index.spec.js
```diff
@@ -332,57 +336,59 @@ describe('Dropzone', () => {
 
     it('should set proper dragActive state on dragEnter', () => {
       const dropzone = mount(
-        <Dropzone />
+        <Dropzone {...classConfig} />
       );
       dropzone.simulate('dragEnter', { dataTransfer: { files } });
-      expect(dropzone.state().isDragActive).toEqual(true);
-      expect(dropzone.state().isDragReject).toEqual(false);
+      expect(dropzone.hasClass(classConfig.activeClassName)).toEqual(true);
```
okonet: I'll deprecate classNames soon and this will cause troubles in the future. Could we check props on children?

okonet:
@fritz-c do you think you could make final changes to this PR so I could merge it?

fritz-c:
I ended up finishing the project that used this, so I'm afraid I don't have time to invest into it now.

okonet:
Closing in favor of #430. Thanks for your work on it.

PR #430, opened 2017-05-24
Handle changes to the `accept` attribute during the drag

Tom-Bonnike:
Open source in a nutshell, eh? 😄 #427 
I wasn’t sure how to deal with: `I'll deprecate classNames soon and this will cause troubles in the future. Could we check props on children?`.
I saw that it was done this way in another test so I just copied it.

### DIFF
--- a/src/index.spec.js
+++ b/src/index.spec.js
@@ -372,6 +372,19 @@ describe('Dropzone', () => {
       expect(dropzone.state().isDragReject).toEqual(true);
     });
 
+    it('should set proper dragActive state if accept prop changes mid-drag', () => {
+      const dropzone = mount(
+        <Dropzone accept="image/*" />
+      );
+      dropzone.simulate('dragEnter', { dataTransfer: { files: images } });
+      expect(dropzone.state().isDragActive).toEqual(true);
+      expect(dropzone.state().isDragReject).toEqual(false);
+
+      dropzone.setProps({ accept: 'text/*' });
+      expect(dropzone.state().isDragActive).toEqual(false);
+      expect(dropzone.state().isDragReject).toEqual(true);
+    });
+
     it('should expose state to children', () => {
       const dropzone = mount(
         <Dropzone accept="image/*">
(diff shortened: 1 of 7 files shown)


Tom-Bonnike:
Oh, didn’t see I needed to do the test change elsewhere. Sorry

src/index.spec.js
```diff
@@ -332,44 +336,68 @@ describe('Dropzone', () => {
...
+          {({ isDragActive, isDragReject }) => {
+            if (isDragReject) {
+              return 'Rejected';
+            }
+            if (isDragActive) {
+              return 'Active';
+            }
+            return 'Empty';
+          }}
+        </Dropzone>
```
Tom-Bonnike: Is this fine?

okonet: I'm wondering if we could just render some simple component with those props and then in the test check if it received the proper props. I think this is what we should test not the classNames — they are just implementation detail.

okonet: Somrthing like

```jsx
const MyComponent = () => null
```

so you can query it with enzyme.

Tom-Bonnike:
```javascript
const DummyChildComponent = () => null;

const dropzone = mount(
  <Dropzone {...classConfig}>
    {props => <DummyChildComponent {...props} />}
  </Dropzone>
);
expect(dropzone.find(DummyChildComponent).props().isDragActive).toEqual(true);
expect(dropzone.find(DummyChildComponent).props().isDragReject).toEqual(false);
```
Like this?

okonet: https://github.com/blainekasten/enzyme-matchers#tohaveproppropkeystring-propvalueany

testSetup.js
```diff
@@ -1,4 +1,5 @@
 /* eslint prefer-template: 0 */
+import 'jest-enzyme';
```
Tom-Bonnike:
```
Missing file extension for "jest-enzyme" (import/extensions)
import 'jest-enzyme';
```

? 👀

okonet: That's weird. I'm checking this out.

okonet: Try to get latest master or this commit: 4af134583be5eab3adb898f879879c25fa7ee517

okonet:
Thanks again for your work on this! Appreciate that you jumped in to it.

PR #174, opened 2016-05-11
Allow selection (input or drop) of directories

tmarshall:
closes #134

PR #436, opened 2017-06-05
Allow upload directories (FileSystem API)

barmaglot92:

- [ ] bugfix
- [x] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [x] Not relevant

- [ ] Not relevant

**Summary**

**Other information**

### DIFF
--- a/examples/Directory/Readme.md
+++ b/examples/Directory/Readme.md
@@ -1,0 +1,38 @@
+Dropzone with default properties and displays list of the dropped files.
+
+```
+class Directory extends React.Component {
+  constructor() {
+    super()
+    this.state = { files: [] }
+  }
+
+  onDrop(files) {
+    this.setState({
+      files
+    });
+  }
+
+  render() {
+    return (
+      <section>
+        <div className="dropzone">
+          <Dropzone onDrop={this.onDrop.bind(this)}>
+            <p>Try dropping some files here, or click to select files to upload.</p>
+          </Dropzone>
+        </div>
+        <aside>
+          <h2>Dropped files</h2>
+          <ul>
+            {
+              this.state.files.map((f, key) => <li key={key.toString()}>path: {f.fullPath}<br />filename: {f.name} - {f.size} bytes</li>)
+            }
+          </ul>
+        </aside>
+      </section>
+    );
+  }
+}
+
+<Directory />
+```
--- a/src/getDataTransferItems.js
+++ b/src/getDataTransferItems.js
@@ -1,17 +1,87 @@
-export default function getDataTransferFiles(event) {
-  let dataTransferItemsList = []
-  if (event.dataTransfer) {
-    const dt = event.dataTransfer
-    if (dt.files && dt.files.length) {
-      dataTransferItemsList = dt.files
-    } else if (dt.items && dt.items.length) {
-      // During the drag even the dataTransfer.files is null
-      // but Chrome implements some drag store, which is accesible via dataTransfer.items
-      dataTransferItemsList = dt.items
+function traverseDirectory(entry, path, entriesList) {
+  const reader = entry.createReader()
+  // Resolved when the entire directory is traversed
+  return new Promise(resolveDirectory => {
+    const iterationAttempts = []
+    ;(function readEntries() {
+      // According to the FileSystem API spec, readEntries() must be called until
+      // it calls the callback with an empty array.  Seriously??
+      reader.readEntries(entries => {
+        if (!entries.length) {
+          // Done iterating this particular directory
+          resolveDirectory(Promise.all(iterationAttempts))
+        } else {
+          // Add a list of promises for each directory entry.  If the entry is itself
+          // a directory, then that promise won't resolve until it is fully traversed.
+          iterationAttempts.push(
+            Promise.all(
+              entries.map(entryObj => {
+                if (entryObj.isFile) {
+                  return new Promise(resolve => {
+                    entryObj.file(file => {
+                      file.fullPath = `${path}/${file.name}` // eslint-disable-line no-param-reassign
+                      entriesList.push(file)
+                      resolve(file)
+                    })
+                  })
+                }
+                // DO SOMETHING WITH DIRECTORIES
+                return traverseDirectory(entryObj, `${path}/${entryObj.name}`, entriesList)
+              })
+            )
+          )
+          // Try calling readEntries() again for the same dir, according to spec
+          readEntries()
+        }
+      })
+    })()
+  })
+}
+
+const addFilesFromItems = items => {
+  let entry
+  let item
+  let i
+  let len
+  const results = []
+  const promises = []
+
+  for ((i = 0), (len = items.length); i < len; i += 1) {
+    item = items[i]
+    entry = item.webkitGetAsEntry()
+    if (item.webkitGetAsEntry != null && entry) {
+      if (entry.isFile) {
+        results.push(item.getAsFile())
+      } else if (entry.isDirectory) {
+        promises.push(traverseDirectory(entry, entry.name, results))
+      }
+    } else if (item.getAsFile != null) {
+      if (item.kind == null || item.kind === 'file') {
+        results.push(item.getAsFile())
+      }
     }
-  } else if (event.target && event.target.files) {
-    dataTransferItemsList = event.target.files
   }
-  // Convert from DataTransferItemsList to the native Array
-  return Array.prototype.slice.call(dataTransferItemsList)
+
+  return Promise.all(promises).then(() => results)
+}
+
+export default function getDataTransferItems(evt) {
+  let files
+  let items
+  let dataTransferItemsList = []
+
+  if (evt.dataTransfer) {
+    files = evt.dataTransfer.files
+    if (files.length) {
+      items = evt.dataTransfer.items
+      if (items && items.length && items[0].webkitGetAsEntry != null) {
+        dataTransferItemsList = addFilesFromItems(items)
+      } else {
+        dataTransferItemsList = Array.prototype.slice.call(files)
+      }
+    }
+  } else if (evt.target && evt.target.files) {
+    dataTransferItemsList = Array.prototype.slice.call(evt.target.files)
+  }
+  return Promise.resolve(dataTransferItemsList)
 }
--- a/src/getDataTransferItems.js
+++ b/src/getDataTransferItems.js
@@ -72,13 +72,15 @@ export default function getDataTransferItems(evt) {
 
   if (evt.dataTransfer) {
     files = evt.dataTransfer.files
-    if (files.length) {
-      items = evt.dataTransfer.items
+    items = evt.dataTransfer.items
+    if (files && files.length) {
       if (items && items.length && items[0].webkitGetAsEntry != null) {
         dataTransferItemsList = addFilesFromItems(items)
       } else {
         dataTransferItemsList = Array.prototype.slice.call(files)
       }
+    } else if (items && items.length) {
+      dataTransferItemsList = Array.prototype.slice.call(items)
     }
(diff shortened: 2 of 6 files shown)


barmaglot92:
Hi, i introduced new FS api for directories upload. getDataTransferItems now return Promise<File[]> intead of File[]. Wait for comments.

okonet:
It seems tests are failing and by looking at the code there is no coverage for it. Do you mind updating existing and adding new testet for this feature?

barmaglot92:
@okonet 
I don`t write tests now becouse i want to discuss way to implement this feature before.
I try to save current way of getting all files at once but alternative way to insert the logic of getDataTransferItems function into Dropzone class. For example we able to call addFile method for adding new File in FileSystemEnty.file callback and render() will be called for every new file.. It will allow to render preview asynchronously for every file (If we need to scan directory)
But i`m not sure with this solution.

barmaglot92:
I'll write the tests for current realization soon

okonet:
Can you show me an example of this async preview generation in code please. I think the direction is good but wondering how this will look like.

barmaglot92:
One question about unit test for our component:
when we simulate dragEnter event

```
const child = dropzone.find(DummyChildComponent)
dropzone.simulate('dragEnter', { dataTransfer: { files } })
expect(child).toHaveProp('isDragActive', false)
expect(child).toHaveProp('isDragReject', true)
```

wa expect that child will has isDragActive but becouse getTransferItems return a promise setState() will be called later:
```
  onDragEnter(evt) {
    evt.preventDefault()

    // Count the dropzone and any children that are entered.
    if (this.dragTargets.indexOf(evt.target) === -1) {
      this.dragTargets.push(evt.target)
    }

    const itemsPromise = getDataTransferItems(evt)

    itemsPromise.then(data => {
      this.setState({ draggedFiles: data })
    })

    if (this.props.onDragEnter) {
      this.props.onDragEnter.call(this, evt)
    }
  }
```
Is it suitable solution?
```
const child = dropzone.find(DummyChildComponent)
      dropzone.simulate('dragEnter', { dataTransfer: { files: files.concat(images) } })
      setTimeout(() => {
        expect(child).toHaveProp('isDragActive', false)
        expect(child).toHaveProp('isDragReject', true)
      })
```

what do u think?

barmaglot92:
@okonet 
I decide to not implement "I try to save current way of getting all files at once but alternative way to insert the logic of getDataTransferItems function into Dropzone class" bacouse it doesn`t anything well for us now.

UTests implemented, but i wait for you comment about solution.
For instance onDragEnter and onDrop use async getDataTransferItems then i need wait for setState (in promise callback):
```
dropzone.simulate('drop', { dataTransfer: { files } })
      getDataTransferFiles({ dataTransfer: { files } }).then(() => {
        expect(dropRejectedSpy.callCount).toEqual(1)
        expect(dropRejectedSpy.lastCall.args[0]).toEqual([...files])
        dropzone.simulate('drop', { dataTransfer: { files: images } })
        getDataTransferFiles({ dataTransfer: { files: images } }).then(() => {
          expect(dropRejectedSpy.callCount).toEqual(1)
          dropzone.simulate('drop', { dataTransfer: { files: files.concat(images) } })
          getDataTransferFiles({ dataTransfer: { files: files.concat(images) } }).then(() => {
            expect(dropRejectedSpy.callCount).toEqual(2)
            expect(dropRejectedSpy.lastCall.args[0]).toEqual([...files])
          })
        })
      })
```

It seems pretty bad, but i don`t know solution..

src/getDataTransferItems.spec.js
```diff
@@ -19,13 +19,16 @@ const files = [
 ]
 
 describe('getDataTransferFiles', () => {
-  it('should return an array', () => {
+  it('should return an promise with an array', done => {
```
okonet: Please consider using http://facebook.github.io/jest/docs/en/asynchronous.html#promises for tests here.

examples/Directory/Readme.md
```diff
@@ -0,0 +1,38 @@
...
+        <div className="dropzone">
+          <Dropzone onDrop={this.onDrop.bind(this)}>
+            <p>Try dropping some files here, or click to select files to upload.</p>
+          </Dropzone>
+        </div>
+        <aside>
+          <h2>Dropped files</h2>
+          <ul>
+            {
+              this.state.files.map((f, key) => <li key={key.toString()}>path: {f.fullPath}<br />filename: {f.name} - {f.size} bytes</li>)
```
okonet: Please use `file` instead of `f` here

examples/Directory/Readme.md
```diff
@@ -0,0 +1,38 @@
+Dropzone with default properties and displays list of the dropped files.
```
okonet: This should probably explain better what this example is demonstrating

src/getDataTransferItems.js
```diff
@@ -1,17 +1,89 @@
...
+        } else {
+          // Add a list of promises for each directory entry.  If the entry is itself
+          // a directory, then that promise won't resolve until it is fully traversed.
+          iterationAttempts.push(
+            Promise.all(
+              entries.map(entryObj => {
+                if (entryObj.isFile) {
+                  return new Promise(resolve => {
+                    entryObj.file(file => {
+                      file.fullPath = `${path}/${file.name}` // eslint-disable-line no-param-reassign
```
okonet: What about Windows?

src/getDataTransferItems.js
```diff
@@ -1,17 +1,89 @@
...
-    const dt = event.dataTransfer
-    if (dt.files && dt.files.length) {
-      dataTransferItemsList = dt.files
-    } else if (dt.items && dt.items.length) {
-      // During the drag even the dataTransfer.files is null
-      // but Chrome implements some drag store, which is accesible via dataTransfer.items
-      dataTransferItemsList = dt.items
+
+  if (evt.dataTransfer) {
+    files = evt.dataTransfer.files
```
okonet: `let { files, items } = evt.dataTransfer`?

src/index.js
```diff
@@ -85,7 +85,11 @@ class Dropzone extends React.Component {
       this.dragTargets.push(evt.target)
     }
 
-    this.setState({ draggedFiles: getDataTransferItems(evt) })
+    const itemsPromise = getDataTransferItems(evt)
```
okonet: Why this assignment is needed? Can't see it being used anywhere.

src/index.js
```diff
@@ -127,7 +131,7 @@ class Dropzone extends React.Component {
 
   onDrop(evt) {
     const { onDrop, onDropAccepted, onDropRejected, multiple, disablePreview, accept } = this.props
-    const fileList = getDataTransferItems(evt)
+    const itemsPromise = getDataTransferItems(evt)
```
okonet: Same here, seems the assignment isn't needed?

src/index.js
```diff
@@ -89,11 +89,11 @@ class Dropzone extends React.Component {
...
     itemsPromise.then(data => {
       this.setState({ draggedFiles: data })
-
-      if (this.props.onDragEnter) {
-        this.props.onDragEnter.call(this, evt)
-      }
     })
+
+    if (this.props.onDragEnter) {
+      this.props.onDragEnter.call(this, evt)
```
barmaglot92: But now don`t sends any param to cb func `this.props.onDragEnter.call(this, evt)` ..

src/index.js
barmaglot92: You are right, i will fix it

src/getDataTransferItems.js
barmaglot92: What are windows in this context ?

src/getDataTransferItems.js
barmaglot92: yes..

examples/Directory/Readme.md
barmaglot92: i remove styleguide point for this feature because it work at all cases

src/getDataTransferItems.js
okonet: Windows Operating System :) Or is this just a name that is used internally?

src/index.js
okonet: Why is that?

src/getDataTransferItems.js
barmaglot92: It`s path to file that i generate by hand, it`s not depends on OS)

okonet: Okay then.

okonet:
One concern about integrating this into the library is that this is only supported in Chrome. And since the library targets multiple browsers and platforms, I'm not sure this is appropriate behavior.

Also I remember there was a PR that was trying implementing this but when decided to go with a standalone library that could be then integrated with react-dropzone. https://github.com/okonet/react-dropzone/pull/174

What do you think?

src/index.spec.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import { mount, render } from 'enzyme'
 import { spy, stub } from 'sinon'
+import getDataTransferFiles from './getDataTransferItems'
```
barmaglot92: Dropzone.onDragEnter now work with async getDataTransferItems, how can i mock it with sync variant in test?)

barmaglot92:
Although this API is experimental, it works in all modern browser now ex (EDGE, FF), but i implement it with checking if browser support it (webitGetAsEntry exists) otherwise in will be works like before

okonet:
If it works in all modern browsers then it's definitely a good addition to this package!

src/index.spec.js
okonet: You're right. What about Jest helpers or async/await? Will this help?

quarklemotion:
FYI, I'm working on a parallel effort to integrate folder uploads with react-dropzone.  I've been working on the https://www.npmjs.com/package/html5-file-selector module and now have it published to npm with some updates.  The code is very similar to what has been added here, but does have some improvements based on several months of production usage and this module has it's own suite of tests, which require a substantial amount of test setup code.  I'm happy to collaborate if you want to try using the html5-file-selector module in this PR.  I originally created that module after attempting a similar react-dropzone PR a while back and we realized that this code could be valuable for other projects and maybe was not best directly incorporated in react-dropzone.

src/getDataTransferItems.js
quarklemotion: I attempted something similar in html5-file-selector (manually setting the `file.fullPath` to the actual nested path), but ran into errors with immutability of File objects with one browser - can't remember which one.  So I ended up wrapping the File object in a wrapper so that this data could be reliably passed back to the client code.

barmaglot92:
@quarklemotion Hi.
I see it also return promise. Obviously we must use your package, but common problems were with test Dropzone Component with setState inside promise callback:
```
dropzone.simulate('drop', { dataTransfer: { files } })
      getDataTransferFiles({ dataTransfer: { files } }).then(() => {
        expect(dropRejectedSpy.callCount).toEqual(1)
        expect(dropRejectedSpy.lastCall.args[0]).toEqual([...files])
        dropzone.simulate('drop', { dataTransfer: { files: images } })
        getDataTransferFiles({ dataTransfer: { files: images } }).then(() => {
          expect(dropRejectedSpy.callCount).toEqual(1)
          dropzone.simulate('drop', { dataTransfer: { files: files.concat(images) } })
          getDataTransferFiles({ dataTransfer: { files: files.concat(images) } }).then(() => {
            expect(dropRejectedSpy.callCount).toEqual(2)
            expect(dropRejectedSpy.lastCall.args[0]).toEqual([...files])
          })
        })
      })
```

I played with jest.runAllTimers/Ticks but didn`t get result

okonet:
It would be great to integrate with this module! 

Can you describe the problem in more details so I can take a look?

PR #437, opened 2017-06-06
Bugfix for issue #252, onClick event being called in input twice. 

kylepotts:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Not relevant

**Summary**
This fixes https://github.com/okonet/react-dropzone/issues/252

Nope no breaking change unless someone took advantage of it being called twice.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -323,7 +323,15 @@ class Dropzone extends React.Component {
       type: 'file',
       style: { display: 'none' },
       multiple: supportMultiple && multiple,
-      ref: el => this.fileInputEl = el, // eslint-disable-line
+      ref: el => {
+        this.fileInputEl = el
+        this.fileInputEl.onclick = event => {
+          event.stopPropagation()
+          if (inputProps.onClick) {
+            inputProps.onClick()
+          }
+        }
+      },
       onChange: this.onDrop
     }
 
--- a/src/index.js
+++ b/src/index.js
@@ -32,6 +32,7 @@ class Dropzone extends React.Component {
     this.onDrop = this.onDrop.bind(this)
     this.onFileDialogCancel = this.onFileDialogCancel.bind(this)
     this.setRef = this.setRef.bind(this)
+    this.onFileInputElClick = this.onFileInputElClick.bind(this)
     this.isFileDialogActive = false
     this.state = {
       draggedFiles: [],
@@ -48,6 +49,7 @@ class Dropzone extends React.Component {
       document.addEventListener('dragover', Dropzone.onDocumentDragOver, false)
       document.addEventListener('drop', this.onDocumentDrop, false)
     }
+    this.fileInputEl.addEventListener('click', this.onFileInputElClick, false)
     // Tried implementing addEventListener, but didn't work out
     document.body.onfocus = this.onFileDialogCancel
   }
@@ -58,6 +60,7 @@ class Dropzone extends React.Component {
       document.removeEventListener('dragover', Dropzone.onDocumentDragOver)
       document.removeEventListener('drop', this.onDocumentDrop)
     }
+    this.fileInputEl.removeEventListener('click', this.onFileInputElClick, false)
     // Can be replaced with removeEventListener, if addEventListener works
     document.body.onfocus = null
   }
@@ -196,6 +199,13 @@ class Dropzone extends React.Component {
     }
   }
 
+  onFileInputElClick(evt) {
+    evt.stopPropagation()
+    if (this.props.inputProps && this.props.inputProps.onClick) {
+      this.props.inputProps.onClick()
+    }
+  }
+
   onFileDialogCancel() {
     // timeout will not recognize context of this method
     const { onFileDialogCancel } = this.props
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -323,7 +323,15 @@ class Dropzone extends React.Component {
       type: 'file',
       style: { display: 'none' },
       multiple: supportMultiple && multiple,
-      ref: el => this.fileInputEl = el, // eslint-disable-line
+      ref: el => {
```
okonet: Can you please extract this to a instance method to prevent from creating a new function each time?

src/index.js
```diff
@@ -323,7 +323,15 @@ class Dropzone extends React.Component {
...
+        this.fileInputEl = el
+        this.fileInputEl.onclick = event => {
```
okonet: I'm wondering if this can break some code that uses even handler on this ref. Also, this will create a memory leak when element is unmounted. Please use `addEventListener` and `removeEventListener` for this.

okonet:
Thank for creating this PR! I'm trying to fix CI at the moment so after some changes and green tests this can be merged.

Does this change need tests be updated? Why are tests not failing now BTW?

src/index.js
kylepotts: Good point. I'll update that.

src/index.js
kylepotts: Can do.

kylepotts:
@okonet I updated that PR for ya. 

I didn't look through all the tests but I can but currently tests fail because of a linting issue. 
https://github.com/okonet/react-dropzone/blob/master/src/index.spec.js#L834

```
/Users/kyle.potts1/Downloads/gh/react-dropzone/src/index.spec.js
  834:32  error  Delete `·(`      prettier/prettier
  846:18  error  Delete `⏎····)`  prettier/prettier
```

I can fix that in this PR too if you'd like.

okonet:
Prettier should run in a pre-commit hook so those should be fixed automatically. Not sure why are they still failing.

src/index.js
```diff
@@ -323,15 +333,7 @@ class Dropzone extends React.Component {
...
-      ref: el => {
-        this.fileInputEl = el
-        this.fileInputEl.onclick = event => {
-          event.stopPropagation()
-          if (inputProps.onClick) {
-            inputProps.onClick()
-          }
-        }
-      },
+      ref: el => this.fileInputEl = el, // eslint-disable-line
```
okonet: Can this also go into a class method, please? Something called `this.setRefs`

src/index.js
```diff
@@ -196,6 +199,13 @@ class Dropzone extends React.Component {
     }
   }
 
+  onFileInputElClick(evt) {
+    evt.stopPropagation()
+    if (this.props.inputProps && this.props.inputProps.onClick) {
+      this.props.inputProps.onClick()
```
okonet: Codecov says this line is uncovered by tests. Mind adding a test for it please?

src/index.js
```diff
@@ -196,6 +199,13 @@ class Dropzone extends React.Component {
     }
   }
 
+  onFileInputElClick(evt) {
```
okonet: I'd name it `onInputElementClick` instead.

kylepotts:
@okonet PR updated and test case added. Everything looks green on my end.

src/index.js
```diff
@@ -48,6 +50,7 @@ class Dropzone extends React.Component {
       document.addEventListener('dragover', Dropzone.onDocumentDragOver, false)
       document.addEventListener('drop', this.onDocumentDrop, false)
     }
+    this.fileInputEl.addEventListener('click', this.onInputElementClick, false)
```
okonet: I'm wondering now why you had to use refs to attache event listener? Was `onClick` in the `inputProps` not enough?

kylepotts: I needed to extend on onClick of the input element to stop event propagation.

okonet: But I still don't get why you have chosen to use refs and DOM API instead of React's one.

Something like this didn't work?

```jsx
{
  ...inputProps,
  onClick: (evt) => { 
    evt.stopPropagation();
    if (typeof inputProps.onClick === 'function') {
      inputProps.onClick()
    }
  }
}
```

kylepotts: I'll give that way a try and report back.

kylepotts: I tried this 

```
    return (
      <div
        className={className}
        style={appliedStyle}
        {...divProps /* expand user provided props first so event handlers are never overridden */}
        onClick={this.onClick}
        onDragStart={this.onDragStart}
        onDragEnter={this.onDragEnter}
        onDragOver={this.onDragOver}
        onDrop={this.onDrop}
        ref={this.setRef}
      >
        {this.renderChildren(children, isDragActive, isDragReject)}
        <input
          {...{
            ...inputProps,
            onClick: evt => {
              evt.stopPropagation()
              if (typeof inputProps.onClick === 'function') {
                inputProps.onClick()
              }
            }
          }}
          {...inputAttributes}
        />
      </div>
    )
```

However I still got the double click, and the test I created for this also failed.

okonet: I'm wondering why. It should not as far as I understand. I thought that's a React bug but I could not reproduce it in isolation. Ideas?

kylepotts: @okonet https://github.com/facebook/react/issues/1691

okonet: Amazing! Thanks for the research!

okonet:
We should re-think this as https://github.com/facebook/react/issues/1691 is closed

okonet:
@kylepotts could you please rebase it on latest master. It should not cause any conflicts but for some reason I don't see the button to do this in the UI.

kylepotts:
@okonet Merged latest master in should be good now.

okonet:
Thanks for this!

PR #451, opened 2017-06-27
Adjust timing around onClick/onDrag* events in IE11/Edge

### LINKED ISSUE #450: IE11/Edge timing issues between `onClick` and `onDrop*` events.

- [x] I found a bug

in IE11/Edge, the `onClick` prop is fired only after the user closes the dialog window after either a selection or a cancel.  

For parity with other browsers, the `onClick` prop should be invoked prior to opening the dialog via `fileInputEl.click()`

1.  create a Dropzone with the following props, 

```jsx
<div>
  <Dropzone
    onClick={() => this.setState({error: null})}
    onDropRejected={() =>this.setState({error: 'you messed up'})}
    maxFilesize="1"
  />
  {this.state.error}
</div>
```

2. Click the drop zone area, select an invalid file from the browser dialogue.

3. In Edge and IE11, the error will not show

The error should show as it does in every other browser.

It seems that in IE11/Edge the `click()` call of a file input will block until after the file dialog has been closed by the user, either by selecting a file or pressing cancel.  This creates a timing issue where the `onClick` fires around the same time as the `onDrop*` events. 

I believe this is react internals at play - likely batching any updates that occur and recognizing the state hasn't changed when it goes to re-render, so in IE/Edge the component is never re-rendered.  In other browsers, it doesn't block and `onClick` fires immediately, regardless of the state of the browse dialogue.

I've been playing around with this in prep for a PR and have noted that even if `onClick` fires before `fileInputEl.click()`, the change won't be shown until after the dialogue is closed.  I think, again, this is react internals at play and it hasn't had time to change the dom before IE/Edge blocks on the dialogue.

I was able to get it to work cleanly by placing the `onClick` invocation before everything and wrapping the `this.open` call in a `setImmediate`  -- (aside, TIL the IE11/Edge/Node are the only runtimes that support setImmediate.  Probably best to use `setTimeout` with a 0 timeout.)

tswaters:
Fixes #450

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

^^ had to fix a few tests to be async with the introduction on a `setTimeout` in the click handler.

- [x] Not relevant

**Summary**

Please see the issue, #450 for most of the reasoning.  Executive summary:  

In IE11/Edge the file browser dialogue will block code running.  A timeout is necessary here so that `setState` calls in `onClick` and `onDragRejected` won't conflict when batched together.

No.

**Other information**

There was an unrelated code style fix that sorta just happened -- see the very bottom of this diff.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -193,10 +193,15 @@ class Dropzone extends React.Component {
     const { onClick, disableClick } = this.props
     if (!disableClick) {
       evt.stopPropagation()
-      this.open()
+
       if (onClick) {
         onClick.call(this, evt)
       }
+
+      // in IE11/Edge the file-browser dialog is blocking, ensure this is behind setTimeout
+      // this is so react can handle state changes in the onClick prop above above
+      // see: https://github.com/okonet/react-dropzone/issues/450
+      setTimeout(this.open.bind(this), 0)
     }
   }
 
--- a/package.json
+++ b/package.json
@@ -52,7 +52,8 @@
   "contributors": [
     "Andrey Okonetchnikov <andrey@okonet.ru> (http://okonet.ru)",
     "Mike Olson <me@mwolson.org>",
-    "Param Aggarwal"
+    "Param Aggarwal",
+    "Tyler Waters <tyler.waters@gmail.com>"
   ],
   "license": "MIT",
   "peerDependencies": {
--- a/src/index.spec.js
+++ b/src/index.spec.js
@@ -175,11 +175,14 @@ describe('Dropzone', () => {
   })
 
   describe('onClick', () => {
-    it('should call `open` method', () => {
+    it('should call `open` method', done => {
       const dropzone = mount(<Dropzone />)
       spy(dropzone.instance(), 'open')
       dropzone.simulate('click')
-      expect(dropzone.instance().open.callCount).toEqual(1)
+      setTimeout(() => {
+        expect(dropzone.instance().open.callCount).toEqual(1)
+        done()
+      }, 0)
     })
 
     it('should not call `open` if disableClick prop is true', () => {
@@ -189,13 +192,16 @@ describe('Dropzone', () => {
       expect(dropzone.instance().open.callCount).toEqual(0)
     })
 
-    it('should call `onClick` callback if provided', () => {
+    it('should call `onClick` callback if provided', done => {
       const clickSpy = spy()
       const dropzone = mount(<Dropzone onClick={clickSpy} />)
       spy(dropzone.instance(), 'open')
       dropzone.simulate('click')
-      expect(dropzone.instance().open.callCount).toEqual(1)
-      expect(clickSpy.callCount).toEqual(1)
+      setTimeout(() => {
+        expect(dropzone.instance().open.callCount).toEqual(1)
+        expect(clickSpy.callCount).toEqual(1)
+        done()
+      }, 0)
     })
 
     it('should reset the value of input', () => {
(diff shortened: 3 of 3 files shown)


tswaters:
Welp.  Turns out those unrelated style fixes that happened automatically broken travis....

src/index.js
```diff
@@ -193,10 +193,14 @@ class Dropzone extends React.Component {
     const { onClick, disableClick } = this.props
     if (!disableClick) {
       evt.stopPropagation()
-      this.open()
+
       if (onClick) {
         onClick.call(this, evt)
       }
+
+      setTimeout(() => {
```
okonet: Could you please add a comment referring to the issue and describing why this is needed?

okonet: Can this be rewritten as `setTimeout(this.open, 0)` you think?

tswaters:
Sure.  The formatting fail was due to using `npm` instead of `yarn` for the install.... the automagical `eslint --fix` did it in a pre-commit hook.  I've fixed up the rest of it.

src/index.js
```diff
@@ -193,10 +193,15 @@ class Dropzone extends React.Component {
...
-      this.open()
+
       if (onClick) {
         onClick.call(this, evt)
       }
+
+      // in IE11/Edge the file-browser dialog is blocking, ensure this is behind setTimeout
+      // this is so react can handle state changes in the onClick prop above above
+      // see: https://github.com/okonet/react-dropzone/issues/450
+      setTimeout(this.open(), 0)
```
okonet: Hmm, I think this isn't the semantic and this as `() => this.open()`. I meant to pass the reference to the function here.

tswaters: Oops!  Yea, you're right that is totally incorrect.... I'll fix this shortly.

tswaters:
Dangit.  I'll look into this CI failure shortly... in the middle of other things.  Stay tuned!

okonet:
Thanks for working on this!

tswaters:
No problem, thanks for the review!

beckend:
This PR and line https://github.com/react-dropzone/react-dropzone/blob/8dd2affcec0206b31bf6dccaba31bf5f422661f8/src/index.js#L204

broke onclick for safari.
Removing the settimeout fixed it for safari.

mrharel:
I have the same issue as @beckend, on mobile iOS the user needs to click double tap on the drop zone to open the file system. removing the `setTimeout` fix it. can somebody check it?

beckend:
@mrharel because i don't really know how to repro it, i decided to workaround the issue, so i check i made a workaround:

```javascript
/**
 * Bug with react-dropzone where onClick does not fire on safari correctly
 * https://github.com/react-dropzone/react-dropzone/issues/502
 * This component was made to fix that
 */
import PropTypes from 'prop-types';
import React, { Component } from 'react';
import ReactDropzone from 'react-dropzone';

export default class Dropzone extends Component {

  static mapBlockingFileDialogBrowsers = {
    EDGE: true,
    MSIE: true
  };

  static propTypes = {
    ...ReactDropzone.propTypes,
    browserName: PropTypes.string.isRequired,
    // eslint-disable-next-line react/forbid-prop-types
    rootProps: PropTypes.any
  };

  handleSetRefDropzone = (ref) => {
    this.refDropzone = ref;
  }

  // basically a modified clickhandler from react-dropzone
  handleOnClick = (evt) => {
    const {
      browserName,
      disableClick,
      onClick
    } = this.props;

    if (disableClick) { return; }

    evt.stopPropagation();

    if (onClick) {
      onClick.call(this.refDropzone, evt);
    }

    if (this.constructor.mapBlockingFileDialogBrowsers[browserName]) {
      // in IE11/Edge the file-browser dialog is blocking, ensure this is behind setTimeout
      // this is so react can handle state changes in the onClick prop above above
      // see: https://github.com/react-dropzone/react-dropzone/issues/450
      setTimeout(this.refDropzone.open.bind(this.refDropzone), 0);

    } else {
      this.refDropzone.open();
    }
  }

  render() {
    // eslint-disable-next-line no-unused-vars
    const { browserName, disableClick, onClick, rootProps, ...passedProps } = this.props;

    return (
      <div {...rootProps} onClick={this.handleOnClick}>
        <ReactDropzone {...passedProps} disableClick ref={this.handleSetRefDropzone} />
      </div>
    );
  }

}
```

mrharel:
@beckend or you can just go back to version `react-dropzone": "3.13.3"` before this fix was done, and see if someone will fix it soon. 

Thanks for sharing this code, i will give it a try.

Issue #321, opened 2017-01-11
Custom validators


okonet:
react-dropzone provides some simple validation out of the box like `accept`, `minSize` and `maxSize` but it lacks rejection reason (#257 and #319) so it's impossible to react to this rejection without the duplicating the validations in the user's component's code.

I have been thinking about implementing some kind of custom validator functions that you could then compose. Each such validator function should have a generic API. Something like this:

```jsx
import { validate, assertMaxFileSize, assertMinFileSize, asserFileType } from 'react-dropzone'

function onValidate(files) {
  const { acceptedFiles, rejectedFiles } = validate(assertMaxFileSize, assertMinFileSize, asserFileType)(files)
  return [ acceptedFiles, rejectedFiles ]
}

const MyDropZone = (props) => {
  return <Dropzone onValidate={onValidate} />
}
```

This would allow create custom validators and react to rejections appropriately in the UI.

Thoughts?

kandros:
I don't know, I believe simple validation like file-size and file-type is better as it is now, handled by the library and based on props, for more complex validation you can already handle all of them in the onDrop function, where you have access  to the files array already.

The example is not actually exposing the failure reasons.

what showed in #257 with the reasons exported (to match rejected files with a filter function) looks more intuitive

okonet:
The example is simplified of course, and the use case with the rejection reason should be handled by it as well (I open for suggestions here). I'm afraid that simple validations are often not enough and I want to keep the surface API of the library minimal.

I realize the current approach is a simpler one, but coupling the lib with validation purposes tends to increase the number of props and different combinations of them which IMO leads to complexity.

For instance, https://github.com/enyo/dropzone is very popular but it tries to solve all use cases which leads to enormous amount of options: http://www.dropzonejs.com/#configuration-options

kandros:
oh I see now, i tried to implement something and totally came up with the same solution, basically `onValidate` will pass `[ acceptedFiles, rejectedFiles ]` to the current implementation of onDrop right?

okonet:
Yeah, this is what I was thinking about. What is your experience with the similar implementation?

kandros:
Looks good to me, passing  `[ acceptedFiles, rejectedFiles ]` instead of a list of validation functions  makes it easier to handle errors or a per file basis instead of relying on the library

what about array of errors to each rejectedFile? so that we can handle multiple errors on a single file and push our internally
[...]

PR #468, opened 2017-08-01
Remove styles and make the component headless

okonet:
🚧🚧🚧
Please don't merge yet. Still WIP!!!
🚧🚧🚧

The idea of this PR is to remove the internal styles and element management and make the library "headless" i.e. not concerned about how styles applied to elements (like downshift). The dropzone component should only be concerned with drag-n-drop behavior leaving style management up to the user since every implementation can be unique.

~~We might consider creating a SimpleDropzone component for the backward compatibility and drop-in replacement for current installations.~~

- [ ] bugfix
- [x] feature
- [x] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [ ] Not relevant

**Summary**

See discussion around https://github.com/okonet/react-dropzone/issues/321

Yes!

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,15 +20,76 @@ Install it from npm and include it in your React build process (using [Webpack](
 npm install --save react-dropzone
 ```
 
-## Usage
-
-Import `Dropzone` in your React component:
+## Usage with ES2015 classes
 
 ```javascript
+import React from 'react'
+import cx from 'classnames'
 import Dropzone from 'react-dropzone'
+
+class MyDropzone extends React.Component {
+
+  onDrop = (acceptedFiles, rejectedFiles) => {
+    // Do something with files
+    // i.e. pre-process, upload etc.
+  }
+
+  render() {
+    return (
+      <Dropzone onDrop={this.onDrop}>
+        {
+          ({ isDragActive }) => {
+            return (
+              <div className={cx('dropzone', {
+                'dropzone--isActive': isDragActive
+              })}>
+              {
+                isDragActive ?
+                  <p>Drop files here...</p> :
+                  <p>Try dropping some files here, or click to select files to upload.</p>
+              }
+              </div>
+            )
+          }
+        }
+      </Dropzone>
+    );
+  }
+}
+
 ``` 
+
+## Usage as a HOC with stateless components
+
+```javascript
+import React from 'react'
+import cx from 'classnames'
+import { makeDropzone } from 'react-dropzone'
+
+function MyDropzone({ isDragActive }) {
+  const classNames = cx('dropzone', {
+    isActive: isDragActive
+  })
+  return (
+    <div className={classNames}>
+      { isDragActive ?
+          <p>Drop files here...</p> :
+          <p>Try dropping some files here, or click to select files to upload.</p>
+      }
+    </div>
+  );
+}
+
+// Call HoC with your component and options as arguments
+export default makeDropzone(MyDropzone, {
+  onDrop: (acceptedFiles, rejectedFiles) => {
+    // Do something with files
+    // i.e. pre-process, upload etc.
+  } 
+})
+```
   
-  and specify the `onDrop` method that accepts two arguments. The first argument represents the accepted files and the second argument the rejected files.
+The `onDrop` method that accepts two arguments. The first argument represents the accepted files and the second argument the rejected files.
   
 ```javascript
 function onDrop(acceptedFiles, rejectedFiles) {
@@ -45,13 +106,13 @@ Using `react-dropzone` is similar to using a file form field, but instead of get
 Specifying the `onDrop` method, provides you with an array of [Files](https://developer.mozilla.org/en-US/docs/Web/API/File) which you can then send to a server. For example, with [SuperAgent](https://github.com/visionmedia/superagent) as a http/ajax library:
 
 ```javascript
-    onDrop: acceptedFiles => {
-        const req = request.post('/upload');
-        acceptedFiles.forEach(file => {
-            req.attach(file.name, file);
-        });
-        req.end(callback);
-    }
+function onDrop(acceptedFiles) {
+  const req = request.post('/upload')
+  acceptedFiles.forEach(file => {
+      req.attach(file.name, file)
+  })
+  req.end(callback)
+}
 ```
 
 ## PropTypes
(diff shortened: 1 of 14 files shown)


okonet:
@SaraVieira it's still WIP and I need some help on this one if you have time.

headzoo:
I like the idea here, because the first thing I do anytime I use this component is remove the default styles.

One thing I noticed was this bit of code [here](https://github.com/react-dropzone/react-dropzone/blob/e1f30525bf7b3d59efd1bf7bc6e8e2b81012340a/src/index.js#L291):

```js
// Remove custom properties before passing them to the wrapper div element
const customProps = [
  'acceptedFiles',
  'preventDropOnDocument',
  'disablePreview',
  'disableClick',
  'onDropAccepted',
  'onDropRejected',
  'onFileDialogCancel',
  'maxSize',
  'minSize'
]
const divProps = { ...rest }
customProps.forEach(prop => delete divProps[prop])
```

Take a look at how I handle this in other projects, like [here](https://github.com/deskpro/deskpro-components/blob/master/src/Components/Alert.jsx#L68):

```js
<div {...objectKeyFilter(props, Alert.propTypes)}>
```

The general idea is getting a list of keys from `Dropzone.propTypes`  instead of maintaining a separate list in `customProps`. That way you don't have to remember to update `customProps` anytime the component props are changed.

okonet:
@headzoo looks like https://github.com/react-dropzone/react-dropzone/blob/e1f30525bf7b3d59efd1bf7bc6e8e2b81012340a/src/index.js#L291-L303 isn't being used in the branch.

hubgit:
I think the concept of this PR is correct, but I wonder if it might make sense to create a brand new class for the HOC, leaving the old one intact but deprecated.

It would mean some duplicated code, until the old class is removed, but would allow for a cleaner rewrite of the main component.

donysukardi:
I attempted to write the render function the primitive way, i.e. leaving the actual rendering to user, ala [downshift](https://github.com/paypal/downshift) so that we don't render unnecessary div.

```js
<Dropzone onDrop={this.onDrop.bind(this)}>
  {({ rootProps, inputProps, isDragActive, isDragAccept, isDragReject }) => (
    <div {...rootProps}> { /* We can fit styled-components perfectly here */ }
      <input {...inputProps} />
    /div>
  )}
</Dropzone>
```

The code can be found here - https://github.com/donysukardi/react-dropzone/blob/51a24e180cbc94bbbbe6b4363a0edf929ff8fe05/src/index.js#L265-L308

okonet:
@donysukardi I like the idea of stealing ideas from downshift. I already have similar approach in mind but I don't have time to finalize it. Feel free to help! Check out my branch and create a PR against it. It needs tests fixed, too. Thanks!

donysukardi:
Sure. Let me work on that! :D

okonet:
@donysukardi after looking at your PR I don't think it's worth doing TBH. I think it's an unnecessary complexity. Instead, I'd suggest we create a prop on the root `Dropzone` component that would allow passing a pre-assembled `input` that would allow customizations but by default, our `input` would just be rendered inside the container with all required props. Thoughts?

okonet:
Want to help?

donysukardi:
Sure.. How could I help? Fire away 🔥

okonet:
Not sure but a review would be good already. Also let us know what you think of the API

README.md
```diff
@@ -20,38 +20,104 @@ Install it from npm and include it in your React build process (using [Webpack](
...
+gives you maximum flexibility to render what, when, and wherever you like. You
+call these on the element in question (for example: `<div {...getRootProps()}`)). It's advisable to pass all your props to that function
+rather than applying them on the element yourself to avoid your props being
+overridden (or overriding the props returned). For example:
+`getRootProps({onClick(event) {console.log(event)}})`.
+
+| property          | type           | description                                                                    |
+| ----------------- | -------------- | ------------------------------------------------------------------------------ |
+| `getRootProps`    | `function({})` | Returns the props you should apply to the root drop container you render       |
+| `getInputProps`   | `function({})` | Returns the props you should apply to hidden file input you render             |
```
bebraw: I think we might be able to generate the tables with markdown-magic. Nicer to maintain.

examples/Basic/Readme.md
```diff
@@ -1,12 +1,15 @@
...
 class Basic extends React.Component {
-  constructor() {
-    super()
+  constructor(props) {
+    super(props)
     this.state = { files: [] }
+    this.onDrop = this.onDrop.bind(this)
   }
-
+
```
bebraw: How new style do you prefer here?

examples/Styling/Readme.md
```diff
@@ -1,23 +1,83 @@
...
+      </div>
+    )
   }}
 </Dropzone>
 ```
+
+### Using styled-components
+
+```
+const styled = require('styled-components').default;
```
bebraw: Maybe ES6 import instead?

README.md
okonet: Ideally these should be generated from code. If that's not possible then yes, let's use some tool. maintaining this will be madness!

examples/Basic/Readme.md
okonet: What?

examples/Styling/Readme.md
okonet: We need to upgrade styleguidist then?

examples/Basic/Readme.md
bebraw: I mean, code style (state to class, new style binding).

examples/Styling/Readme.md
bebraw: Likely worth doing in a separate PR, yeah. No need to fix yet.

src/index.js
```diff
@@ -400,9 +333,22 @@ Dropzone.propTypes = {
   accept: PropTypes.string,
 
   /**
-   * Contents of the dropzone
+   * Render function that renders the actual component. It receives object with the following properties,
+   * @param {Function} getRootProps
```
rolandjitsu: I think the jsdoc syntax for [params](http://usejsdoc.org/tags-param.html) is:
```js
/**
 * @param {Object} props
 * @param {Function} props.getInputProps
 ...
 */
```

Issue #472, opened 2017-08-02
Feature request: disable


maasha:
It would be grand if there was an option to disable the dropzone entirely, so that it only renders, but has no functionality. That is useful for tasks done in a particular order - to have the dropzone disabled until some configuration steps on the webpage are complete.

okonet:
Great feature request. Do you want to work on this?

maasha:
I have limited experience with Javascript and have no insight into how react-dropzone works - it should surely be something quick to do for someone who have her mind wrapped around the code already. I'd be happy to test.

kainanaina:
Where can I find information about minimal browser support?

trinhngocdieu:
[Zalo Chat](http://chat.zalo.me)

okonet:
@kainanaina how is your question related to this issue?

okonet:
Thanks @oieduardorabelo for working on this! This is now implemented in https://github.com/react-dropzone/react-dropzone/pull/473

PR #473, opened 2017-08-03
Feature: Add disable prop

oieduardorabelo:

- [ ] bugfix
- [x] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [ ] Not relevant

**Summary**

feat request in https://github.com/okonet/react-dropzone/issues/472, be able to  to disable the dropzone entirely

hey @okonet , just one idea to implement #472


src/index.js
```diff
@@ -410,6 +424,11 @@ Dropzone.propTypes = {
   disableClick: PropTypes.bool,
 
   /**
+ * Enable/disable the dropzone entirely
+ */
+  disableDropzone: PropTypes.bool,
```
okonet: Let's rename the prop to `disabled` to conform the DOM API: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

src/index.js
```diff
@@ -344,7 +356,8 @@ class Dropzone extends React.Component {
       style: { display: 'none' },
       multiple: supportMultiple && multiple,
       ref: this.setRefs,
-      onChange: this.onDrop
+      onChange: this.onDrop,
+      disabled: disableDropzone
```
okonet: After the prop is renamed, let's put it after the `accept` one using the shorthand notation.

src/index.js
```diff
@@ -66,6 +69,14 @@ class Dropzone extends React.Component {
     document.body.onfocus = null
   }
 
+  composeHandlers(handler, target) {
```
okonet: Since this is only called in the constructor and it returns `null`, this won't react to the change of the prop. We'll need to add a test for this (change from disabled to enabled state and check that events are being fired).

okonet:
Thanks for working on it!

jacksonrayhamilton:
Thanks @oieduardorabelo for adding this!  And just in time, as our product just started needing this functionality, as well.  :smile:

oieduardorabelo:
hey @okonet,

- I've moved the compose fn to the render and added tests for the react lifecycle updates
- I also exposed `disabledStyle` and `disabledClassName`, you know, stylesss! (maybe the default `opacity` isn't a good idea? anyway, up to you)

jacksonrayhamilton:
@okonet Can we expect a release with this feature soon?

okonet:
See https://github.com/react-dropzone/react-dropzone/releases/tag/v4.1.0

It wasn't in /releases since I've moved the repo to another organization and some semantic-release functions were fucked up.

jacksonrayhamilton:
Ah, missed that.  Thanks!

PR #476, opened 2017-08-07
Set autocomplete="off" on input attribute

### LINKED ISSUE #475: Maybe disable auto complete on input element by default

- [x] I found a bug

Firefox throws a SecurityError. I think the reason for this is [described here](https://stackoverflow.com/a/37972457/931860). Basically, I think Firefox tries to set the input element's value which throws the error. When I pass `{ autoComplete: 'off' }` to Dropzone in the `inputProps` the problem goes away.

1. Drop a file or pick one using the file UI
2. Refresh the browser

No error

Firefox 54.0.1 Windows

Thanks for the great component, btw. :)

dominic-p:
Fixes #475.

Note that because `inputAttributes` is placed after user provided `inputProps` when they are merged, users will not be able to override this setting. Since, I can't imagine why someone would want this behavior, this way makes more sense and keeps the code clean.

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Not relevant

**Summary**

**Other information**


okonet:
Thanks!

PR #487, opened 2017-08-30
[WIP] Enabling keyboard navigation / Accessibility fixes

holloway:
This is a Work In Progress and I'd just like feedback on the approach before proceeding.

Rather than the other related issues that suggest a top-level `<button>` this just uses a different technique to hide the input (`width:0, height:0, overflow:hidden`) which is keyboard navigable. It also adds a new focusable style and className override like the other events.

- [ ] bugfix
- [X] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [X] Not yet!

I will add tests if you agree with the basic approach that I've made.

- [X] Not relevant

**Summary**

The existing UI isn't keyboard navigable, and this improves accessibility.

https://github.com/react-dropzone/react-dropzone/issues/161
https://github.com/react-dropzone/react-dropzone/pull/336
https://github.com/react-dropzone/react-dropzone/issues/408

No, it should be a seamless upgrade, although it needs cross-browser testing.

**Other information**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -23,6 +23,8 @@ class Dropzone extends React.Component {
     this.onDragOver = this.onDragOver.bind(this)
     this.onDragStart = this.onDragStart.bind(this)
     this.onDrop = this.onDrop.bind(this)
+    this.onFocus = this.onFocus.bind(this)
+    this.onBlur = this.onBlur.bind(this)
     this.onFileDialogCancel = this.onFileDialogCancel.bind(this)
     this.onInputElementClick = this.onInputElementClick.bind(this)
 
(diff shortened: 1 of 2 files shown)


okonet:
Thanks for working on it.

I'm not sure about the solution, though. First of all, I don't like the introduction of the focus state and styles since this should be managed by the browser and CSS ideally.

I'm wondering if something as simple as adding `tabIndex={0}` to the wrapper component would solve it...

@SaraVieira thoughts?

Konsada:
Was this every resolved?

merric:
all of the tasks for this have been closed automatically. How come no progress here?

rolandjitsu:
@merric there was an attempt in https://github.com/react-dropzone/react-dropzone/pull/700, but we first have to finish https://github.com/react-dropzone/react-dropzone/pull/468.

I might have some time this coming month to continue what @okonet started.

PR #510, opened 2017-09-27
Add es2015 module build

anajavi:

- [ ] bugfix
- [ ] feature
- [ ] refactoring / style
- [X] build / chore
- [ ] documentation

- [ ] Not relevant

- [X] Not relevant

**Summary**

Adds build for es2015 modules. This enables Webpack 3 to do scope hoisting and tree shaking. In my application this saves 3338 bytes before gzipping.

In order to application benefit from this build with webpack, module transformations must be disabled in applications babel configuration. For example:
```
"presets": [["env",{ "modules": false }], "react"]
```

Closes https://github.com/react-dropzone/react-dropzone/issues/496

No.

**Other information**

The .babelrc file has same config three times, this should be fixable when babel 7 is released.


.babelrc
```diff
@@ -1,4 +1,19 @@
 {
-  "presets": ["env", "react", "stage-1"],
-  "plugins": ["add-module-exports"]
+  "env": {
+    "development": {
```
okonet: Can we remove this code duplication by using `es5` and `es6` as `BABEL_ENV` explicitly?

package.json
```diff
@@ -2,17 +2,21 @@
   "name": "react-dropzone",
   "description": "Simple HTML5 drag-drop zone with React.js",
   "main": "dist/index.js",
+  "module": "dist/es/index.js",
   "scripts": {
     "ci": "git-cz",
     "clean": "rimraf ./dist",
-    "build": "npm run clean && webpack",
+    "build": "npm run clean && webpack && npm run build:es",
```
okonet: So here we'd add `BABEL_ENV=es5`

.babelrc
rkmarks: good call out! I missed that on my initial review

anajavi:
The problem with that is that then we have to prepend every script in package.json with either one. This is because if babel does not find BABEL_ENV, it will use NODE_ENV. The only thing babel does for default is to use "development" env if none is specified.

And the second problem that I just checked is, that the variable doesn't pass to second command on line. For example this one causes babel to build with "development" env, which makes commonjs modules:
```
"build:es": "BABEL_ENV=es echo test && babel ./src --out-dir ./dist/es --ignore spec.js"
```

Actually the current master branch has problem too, this line builds with correct NODE_ENV, but jest is run with empty NODE_ENV causing the test target to be in src:
```
"prepublish": "NODE_ENV=production npm run build && jest"
```

I know it is ugly to define the same line several times, but there doesn't seem to better options available until babel 7. Babel 7 has js config in .babelrc.js, where these settings can be merged.

anajavi:
I tested building with babel 7. Here's a config that works with it:
```javascript
//.babelrc.js
const configs = {
  "es5": {
    "presets": ["env", "react", "stage-1"],
    "plugins": ["add-module-exports"]
  },
  "es6": {
    "presets": [["env",{ "modules": false }], "react", "stage-1"]
  }
}

module.exports = function(api) {
  if(api && api.cache) api.cache(false)
  return "es6" === process.env.BABEL_ENV ? configs["es6"] : configs["es5"]
}
```
 
Funnily enough, with Babel 6 the following .babelrc lets us use the new config.
```javascript
{
  "presets": ["./.babelrc.js"]
}
```

However, I had lots of strange caching problems with Babel 6 before arriving with this config. Also, the new config API is not documented anywhere yet.  It is probably too brittle to use now.

package.json
```diff
@@ -2,17 +2,21 @@
...
     "start": "styleguidist server",
     "styleguide": "styleguidist build",
     "test": "npm run eslint:src && jest --coverage && npm run size",
+    "test:es": "JEST_TARGET=../dist/es/index jest --coverage",
+    "test:umd": "JEST_TARGET=../dist/index jest --coverage",
     "eslint:src": "eslint ./src ./examples ./*.js",
     "commitmsg": "commitlint -e",
     "precommit": "lint-staged",
-    "prepublish": "NODE_ENV=production npm run build && jest",
+    "prepublish": "NODE_ENV=production npm run build && npm run test:umd && npm run test:es",
```
okonet: I'm also wondering if we should keep testing compiled targets. I know I'm the person who introduced that but with this PR it makes kind'a awful and I'm not sure it has enough value since Babel should be well-tested. Thoughts?

anajavi: I have recently had my share of broken npm dependencies.

One still had Flow types in transpiled output, other one was published as an application not a commonjs nor umd module. And one had React embedded in output, making it kind of big..

So I am pro target testing at the moment :)

anajavi:
Is there still something to address, or can this be considered complete?

.babelrc
```diff
@@ -1,4 +1,19 @@
...
+      "presets": ["env", "react", "stage-1"],
```
okonet: One question just popped my mind: do we still need `stage-1` when using `env` preset? If so, what for?

package.json
okonet: I'd prefer we have 2 separate `build` scripts: `build:umd` and `build:es` to make the intent clear.

`build:umd` would be `NODE_ENV=production webpack`

package.json
okonet: So here we could drop the `NODE_ENV` thing.

.babelrc
anajavi: I have never used that preset. The documentation of env says:
"Without any configuration options, babel-preset-env behaves exactly the same as babel-preset-latest (or babel-preset-es2015, babel-preset-es2016, and babel-preset-es2017 together)."

I think that the stage-1 is only needed, if there is some code that depend on current stage-1 propositions.

package.json
anajavi: I'll do this and get back to you

.babelrc
okonet: Yeah, I think stage-1 was needed for rest-spread. I'm wondering how env will transpile it, though.

anajavi: env preset does not yet transpile object-rest-spread. I am using the following:
```
plugins: ["transform-object-rest-spread"]
```
I think that makes it more definitive.

PR #511, opened 2017-10-03
Fix: Correct behaviour on multiple=false

### LINKED ISSUE #458: Reject all if multiple prop is false and multiple files are dropped

- [ ] I found a bug

When the `multiple` prop is false, and the user drops multiple files, the first file is accepted and the rest are rejected.

I would expect all images to be rejected in this case. It's bad UI to show a general error when one file (pseudo-randomly) will succeed.

Happy to make a PR for this if you are open to changing the interface.

DeFuex:

- [x] bugfix
- [x] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Not relevant

**Summary**

Corrects the behaviour when multiple=false on a multiple file drop. Due to changes in  ticket #409 only the css was changed accordingly, but there was still some leftover functionality where a single file still could be accepted on a multiple file drop. The solution here presented should solve the problem.

Closes #458 #409 

**Other information**
There are some leftover test cases that i need to update accordingly 😃


src/index.spec.js
```diff
@@ -580,17 +580,19 @@ describe('Dropzone', () => {
...
-      expect(rejected.length).toEqual(0)
-    })
+    // it('should allow single files to be dropped if multiple is false', () => {
+    //   const dropzone = mount(<Dropzone accept="image/*" onDrop={dropSpy} multiple={false} />)
+
+    //   dropzone.simulate('drop', { dataTransfer: { files: [images[0]] } })
+    //   const [accepted, rejected] = dropSpy.firstCall.args
+    //   expect(accepted.length).toEqual(1)
+    //   expect(rejected.length).toEqual(0)
+    // })
```
rkmarks: @DeFuex why is this test commented out?

src/index.js
```diff
@@ -437,8 +440,8 @@ Dropzone.propTypes = {
   disableClick: PropTypes.bool,
 
   /**
- * Enable/disable the dropzone entirely
- */
+   * Enable/disable the dropzone entirely
+   */
```
rkmarks: Can you format these back? seems a bit quirky out of line w/ the prop. Sorry just a nitpick

DeFuex:
is it possible to label this pr with the #hacktoberfest label? 😄

src/index.spec.js
DeFuex: hi, since the multiple=false option shouldn't allow a single file to be uploaded on a multiple file drop this test case is not necessary i guess -> `/* This test is not needed since that behaviour changed */`

src/index.js
DeFuex: sure 😅

src/index.spec.js
```diff
@@ -580,17 +580,19 @@ describe('Dropzone', () => {
...
-    it('should allow single files to be dropped if multiple is false', () => {
-      const dropzone = mount(<Dropzone accept="image/*" onDrop={dropSpy} multiple={false} />)
+    /* This test is not needed since that behaviour changed */
 
-      dropzone.simulate('drop', { dataTransfer: { files: [images[0]] } })
-      const [accepted, rejected] = dropSpy.firstCall.args
-      expect(accepted.length).toEqual(1)
-      expect(rejected.length).toEqual(0)
-    })
+    // it('should allow single files to be dropped if multiple is false', () => {
```
okonet: This should probably just change to the negation to make the new behavior explicit.

okonet:
The code coverage decreased. Also we should probably update the documentation.

src/index.spec.js
DeFuex: you mean like `expect(rejected.length).not.toEqual(2)` ?

src/index.js
okonet: We don't need to format. prettier should take care of this. Did lint-staged run?

src/index.spec.js
okonet: No I mean the whole test needs to be updated accordingly.

src/index.spec.js
```diff
@@ -583,16 +583,16 @@ describe('Dropzone', () => {
       expect(rejected.length).toBeGreaterThan(1)
     })
 
-    /* This test is not needed since that behaviour changed */
-
-    // it('should allow single files to be dropped if multiple is false', () => {
-    //   const dropzone = mount(<Dropzone accept="image/*" onDrop={dropSpy} multiple={false} />)
+    it('should allow single files to be dropped if multiple is false', () => {
```
okonet: This says "it should allow single files" and the PR is "fixing" it.

src/index.spec.js
```diff
@@ -583,16 +583,16 @@ describe('Dropzone', () => {
...
+      )
 
-    //   dropzone.simulate('drop', { dataTransfer: { files: [images[0]] } })
-    //   const [accepted, rejected] = dropSpy.firstCall.args
-    //   expect(accepted.length).toEqual(1)
-    //   expect(rejected.length).toEqual(0)
-    // })
+      dropzone.simulate('drop', { dataTransfer: { files: images[0].accept } })
+      const [accepted, rejected] = dropAcceptedSpy.firstCall.args
+      expect(accepted.length).not.toEqual(1)
```
okonet: `.not.toEqual(1)` => `.toEqual(0)`?

okonet:
Tests are still failing.

src/index.js
okonet: Also, I agree this looks better now so let's keep it, please.

src/index.js
```diff
@@ -174,9 +174,12 @@ class Dropzone extends React.Component {
     })
 
     if (!multiple) {
```
okonet: I'm wondering if this code can be simplified. Also, should we this be the first check and if this doesn't pass, we should return earlier? As far as I understand it `if (!multiple && this.state.draggedFiles.length > 1)` we want to reject the whole drag operation. Correct?

src/index.spec.js
DeFuex: im not sure about the test case here, but should both, accepted and rejected be like this? ``` expect(accepted.length).toEqual(0)
      expect(rejected.length).toEqual(0)```

src/index.js
DeFuex: that would also be my logical understanding. i can try it out though and change it if necessary.

rkmarks:
@DeFuex looks like tests are still failing on your updates

DeFuex:
@rkmarks yep, gonna work on it this week

DeFuex:
i don't know how far related this is, but since i updated my yarn the tests are successful without changing anything. weird...im gonna push this, lets see if it works now

src/index.spec.js
```diff
@@ -580,15 +580,17 @@ describe('Dropzone', () => {
...
       })
       const rejected = dropSpy.firstCall.args[1]
-      expect(rejected.length).toEqual(2)
+      expect(rejected.length).toBeGreaterThan(1)
     })
 
     it('should allow single files to be dropped if multiple is false', () => {
-      const dropzone = mount(<Dropzone accept="image/*" onDrop={dropSpy} multiple={false} />)
+      const dropzone = mount(
+        <Dropzone accept="image/*" onDrop={dropAcceptedSpy} multiple={false} />
```
okonet: Not sure why did you rename the spy name. This is a bit misleading now IMO since it includes both accepted and rejected files.

src/index.spec.js
```diff
@@ -580,15 +580,17 @@ describe('Dropzone', () => {
...
+      )
 
-      dropzone.simulate('drop', { dataTransfer: { files: [images[0]] } })
-      const [accepted, rejected] = dropSpy.firstCall.args
-      expect(accepted.length).toEqual(1)
+      dropzone.simulate('drop', { dataTransfer: { files: images[0].accept } })
+      const [accepted, rejected] = dropAcceptedSpy.firstCall.args
+      expect(accepted.length).toEqual(0)
       expect(rejected.length).toEqual(0)
```
okonet: If the file isn't accepted, when it should be rejected? Why is this test only updates the first value but not the second?

src/index.spec.js
```diff
@@ -580,15 +580,17 @@ describe('Dropzone', () => {
         dataTransfer: { files: images.concat(files) }
       })
       const rejected = dropSpy.firstCall.args[1]
-      expect(rejected.length).toEqual(2)
+      expect(rejected.length).toBeGreaterThan(1)
     })
 
     it('should allow single files to be dropped if multiple is false', () => {
```
okonet: The test name is misleading

src/index.js
okonet: Did you try this out already? Doesn't look like it has been changed.

src/index.spec.js
```diff
@@ -580,15 +580,17 @@ describe('Dropzone', () => {
...
-      const dropzone = mount(<Dropzone accept="image/*" onDrop={dropSpy} multiple={false} />)
+      const dropzone = mount(
+        <Dropzone accept="image/*" onDrop={dropAcceptedSpy} multiple={false} />
+      )
 
-      dropzone.simulate('drop', { dataTransfer: { files: [images[0]] } })
-      const [accepted, rejected] = dropSpy.firstCall.args
-      expect(accepted.length).toEqual(1)
+      dropzone.simulate('drop', { dataTransfer: { files: images[0].accept } })
```
okonet: What is `images[0].accept`? `files` must be an array, right? Why did you change this line?

src/index.js
DeFuex: Working on it now, 'if (!multiple && this.state.draggedFiles.length > 1)' doesn't work, but changing it to 'if (!multiple && this.state.draggedFiles.length === 1)' works, since the length has to be exactly 1 to be accepted as 1 file drop and rejecting multiple file drops', gonna push it when i got my head around the jest test cases.

DeFuex: taking it back, it doesn't work at all, that's why tests are failing, gonna let it as it is

src/index.spec.js
DeFuex: sry, gonna change it back with the next commit.

src/index.spec.js
DeFuex: gonna change the test according to its behaviour

src/index.spec.js
DeFuex: gonna change it back

src/index.spec.js
DeFuex: changing the test name according to its behaviour

src/index.js
```diff
@@ -174,9 +174,12 @@ class Dropzone extends React.Component {
...
-      // if not in multi mode add any extra accepted files to rejected.
-      // This will allow end users to easily ignore a multi file drop in "single" mode.
-      rejectedFiles.push(...acceptedFiles.splice(1))
+      // Only allow one file to be accepted if multiple=false
+      if (this.state.draggedFiles.length === 1) {
+        rejectedFiles.push(...acceptedFiles.splice(1))
+      } else {
+        rejectedFiles.push(...acceptedFiles.splice(0))
```
okonet: Why is this `splice` needed here if you already doing `...`?

src/index.spec.js
```diff
@@ -580,16 +580,16 @@ describe('Dropzone', () => {
         dataTransfer: { files: images.concat(files) }
       })
       const rejected = dropSpy.firstCall.args[1]
-      expect(rejected.length).toEqual(2)
+      expect(rejected.length).toBeGreaterThan(1)
     })
 
-    it('should allow single files to be dropped if multiple is false', () => {
+    it('should reject single files on multiple drop if multiple is false', () => {
```
okonet: Is there also a test where you test that all files would end up in `rejected`?

okonet: Or why do you specifically test for single files here?

DeFuex: cause a single file should be rejected when a multiple drop is done with the multiple boolean set on false. As i understand it that's what is tested now.

To the first comment, i could change `expect(rejected.length).toEqual(1)` to `expect(rejected.length).toBeGreaterThanOrEqual(1)` then every case would be checked here.

src/index.js
DeFuex: i thought the spread operator in javascript just turns items in an array to objects with a `key: value` pair? and as i found out splice used with 1 argument is a little hacky? (not sure though)

src/index.spec.js
DeFuex: nevermind, i added a test case for rejected files on a multiple file drop if multiple=false

src/index.js
okonet: > i thought the spread operator in javascript just turns items in an array to objects with a key: value pair? 

No, this is not what spread is doing: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator

There is no need in `.splice(0)` since it's literally what `...` is doing. I'm not even sure we need `...` in front of the first one — seems like a redundant code to me. Please check this.

duro:
@DeFuex AND @okonet You may want to see my comment here:

https://github.com/react-dropzone/react-dropzone/issues/458#issuecomment-352173234

It may be relevant to the work you are doing here

okonet:
Can someone finalize and merge this?

Schmitze333:
Is this still active? Otherwise, I would get it a try... I'd need the described behavior, too.

okonet:
@Schmitze333 would you consider finilazing this PR or do you want to create a new one?

Schmitze333:
Still don't know yet... since the PR is very old, I am thinking about starting out from the current master. If so, I'll open a new PR and link the relevant issue. I will go over it on Friday and start with some examination. I will decide then.

okonet:
In this case it’s probably wise to branch off https://github.com/react-dropzone/react-dropzone/pull/468

PR #517, opened 2017-10-13
fix: Add null check for node in onDocumentDrop

### LINKED ISSUE #492: onDocumentDrop causes an error in IE
On first drop - all is Ok.
When I drop file second time without page refresh in Internet Explorer 11 - 2 requests are sent: one with an empty file, the second with a normal file.

pcgilday:

- [x] bugfix

- [x] Not relevant

**Summary**

- Protect calling `this.node.contains` when `this.node` is null
- Fix IE11 bug where `onDocumentDrop` gets called with null `this.node` ref

closes #492

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -71,7 +71,7 @@ class Dropzone extends React.Component {
   }
 
   onDocumentDrop(evt) {
-    if (this.node.contains(evt.target)) {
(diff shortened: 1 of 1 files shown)


okonet:
Thanks for working on it!

PR #525, opened 2017-10-25
fix: click on dropzone not working on Safari 10.1

izadmehr:
Change triggering click on the input to label with for attribute

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [x] Not relevant

- [x] Not relevant

**Summary**

The click event does not trigger on Safari 10.1. This is due to a security restriction. 

https://github.com/react-dropzone/react-dropzone/issues/522

No
**Other information**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -252,6 +252,7 @@ class Dropzone extends React.Component {
   setRefs(ref) {
     this.fileInputEl = ref
   }
+
   /**
    * Open system file upload dialog.
    *
@@ -286,6 +287,7 @@ class Dropzone extends React.Component {
       multiple,
       name,
       rejectClassName,
+      disableClick,
       ...rest
     } = this.props
 
@@ -363,12 +365,10 @@ class Dropzone extends React.Component {
       multiple: supportMultiple && multiple,
       ref: this.setRefs,
       onChange: this.onDrop,
-      autoComplete: 'off',
-      id: 'selectFile'
+      autoComplete: 'off'
     }
 
     const labelAttributes = {
-      htmlFor: 'selectFile',
       style: {
         width: '100%',
         height: '100%',
@@ -399,6 +399,15 @@ class Dropzone extends React.Component {
     ]
     const divProps = { ...props }
     customProps.forEach(prop => delete divProps[prop])
+    const childContainer = (
+      <div>
+        <input
+          {...inputProps /* expand user provided inputProps first so inputAttributes override them */}
+          {...inputAttributes}
+        />
+        {this.renderChildren(children, isDragActive, isDragAccept, isDragReject)}
+      </div>
+    )
 
     return (
       <div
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -364,7 +363,18 @@ class Dropzone extends React.Component {
       multiple: supportMultiple && multiple,
       ref: this.setRefs,
       onChange: this.onDrop,
-      autoComplete: 'off'
+      autoComplete: 'off',
+      id: 'selectFile'
```
okonet: What happens if user have multiple dropboxes on a single page?

okonet: We should generate unique IDs here.

src/index.js
```diff
@@ -404,7 +414,9 @@ class Dropzone extends React.Component {
         ref={this.setRef}
         aria-disabled={disabled}
       >
-        {this.renderChildren(children, isDragActive, isDragAccept, isDragReject)}
+        <label {...labelAttributes}>
```
okonet: I'm wondering if `label` should become the main container instead of `div`? Will it have any styling implications?

izadmehr:
The `for` attribute worked just on `label`, we can't change it to `div`. I check it, it adds no more styling. For having more than one drop-zone, the `id` and `for` attribute removed, and input moved to the label as a child.

src/index.js
```diff
@@ -367,6 +368,15 @@ class Dropzone extends React.Component {
       autoComplete: 'off'
     }
 
+    const labelAttributes = {
+      style: {
+        width: '100%',
+        height: '100%',
+        margin: 0,
+        cursor: 'pointer'
```
okonet: Not sure about this one especially without a way to customize it. Please remove cursor styling for now to make it non-breaking change.

src/index.js
```diff
@@ -404,11 +423,7 @@ class Dropzone extends React.Component {
         ref={this.setRef}
         aria-disabled={disabled}
       >
-        {this.renderChildren(children, isDragActive, isDragAccept, isDragReject)}
-        <input
-          {...inputProps /* expand user provided inputProps first so inputAttributes override them */}
-          {...inputAttributes}
-        />
+        {disableClick ? childContainer : <label {...labelAttributes}>{childContainer}</label>}
```
okonet: Is there a way to always render the same HTML but prevent clicking using JS?

okonet:
Some tests are failing. Please update tests so I can test this branch.

rohit-ravikoti:
Hi,
Is this being worked on? We are having the same issues where our users can't select files on ios 10.1 devices.

rkmarks:
It looks to have stalled @rohit-ravikoti if you are familiar w/ the fix you can open a PR as well. @izadmehr are you still working on this? if so can you please fix the failing tests so we can test this branch.

PR #528, opened 2017-11-08
feat: add childrenProps

matttti:

- [ ] bugfix
- [x] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Not relevant

**Summary**

Additionally to internal props like `isDragActive` or the state, i think it is useful to also pass custom props to a children function. In my particular use case, I want to re-use the function in several places, but also want to apply custom labels and styling to the content.

No

**Other information**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -266,8 +266,8 @@ class Dropzone extends React.Component {
   renderChildren = (children, isDragActive, isDragAccept, isDragReject, childrenProps) => {
     if (typeof children === 'function') {
       return children({
-        ...this.state,
         ...childrenProps,
+        ...this.state,
         isDragActive,
         isDragAccept,
         isDragReject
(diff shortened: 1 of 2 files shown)


okonet:
I think this will become obsolete as soon as #468 is done, right?

matttti:
ah yes, indeed. I didn't see #468. Is there plan when it's gonna be incorporated?

However, in my opinion the approach here may be more straightforward for common cases than using HOCs as children.

okonet:
@matttti for some generic cases, probably. But the goal is to be able to make the library as less concerned about the actual UI as possible and let developers decide.

okonet:
I'm closing it in favor of #468

okonet:
@matttti there is no ETA on it since I don't have much time to work on it. Do you want to jump in and do it?

Issue #560, opened 2018-01-22
4.2.7 isFileDialogActive timeout issue


timothyallan:
Heya all.

- [x] I found a bug
- [ ] I want to propose a feature

**What is the current behavior?**
as of 4.2.7 I am getting a null issue in the `onFileDialogCancel` timeout.

In the below snippet, _this4.fileInputEl is null, and so .files fails on it. 4.2.6 works great.I've tested the lastest versions of Chrome, and even IE11

```
 key: 'onFileDialogCancel',
    value: function onFileDialogCancel() {
      var _this4 = this;

      // timeout will not recognize context of this method
      var onFileDialogCancel = this.props.onFileDialogCancel;
      // execute the timeout only if the FileDialog is opened in the browser

      if (this.isFileDialogActive) {
        setTimeout(function () {
          // Returns an object as FileList
          var files = _this4.fileInputEl.files;


          if (!files.length) {
            _this4.isFileDialogActive = false;
          }

          if (typeof onFileDialogCancel === 'function') {
            onFileDialogCancel();
          }
        }, 300);
      }
    }
```
**If the current behavior is a bug, please provide the steps to reproduce.**

I'm selecting a file from the browser file popup, which replaces the dropzone component with an image preview. The error occurs when the preview appears, only in 4.2.7.

**What is the expected behavior?**

No error.

okonet:
Looks like https://github.com/react-dropzone/react-dropzone/pull/556 introduced a regression. Would you be ready to help fixing it? Probably a null check will suffice.

timothyallan:
My heart says yes, my multiple project deadlines say probably not :)

yuri-sakharov:
Hi,
when will be merge or better solution?

PR #561, opened 2018-01-25
 fix: check for null [#560] && fix lint errors

thaddeusreid:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [x] Not relevant

- [x] Not relevant

**Summary**

Per [#560] a TypeError is thrown because `this.fileInputEl.files` is referenced in cases where `this.fileInputEl` is null.

I also fixed 3 eslint errors that address comment indenting/whitespace.

no

**Other information**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -239,7 +239,7 @@ class Dropzone extends React.Component {
     // execute the timeout only if the FileDialog is opened in the browser
     if (this.isFileDialogActive) {
       setTimeout(() => {
-        if (this.fileInputEl) {
+        if (this.fileInputEl != null) {
           // Returns an object as FileList
           const { files } = this.fileInputEl
 
--- a/src/index.js
+++ b/src/index.js
@@ -239,11 +239,13 @@ class Dropzone extends React.Component {
     // execute the timeout only if the FileDialog is opened in the browser
     if (this.isFileDialogActive) {
       setTimeout(() => {
-        // Returns an object as FileList
-        const { files } = this.fileInputEl
+        if (this.fileInputEl) {
+          // Returns an object as FileList
+          const { files } = this.fileInputEl
 
-        if (!files.length) {
-          this.isFileDialogActive = false
+          if (!files.length) {
+            this.isFileDialogActive = false
+          }
         }
 
         if (typeof onFileDialogCancel === 'function') {
@@ -396,7 +398,9 @@ class Dropzone extends React.Component {
       <div
         className={className}
         style={appliedStyle}
-        {...divProps /* expand user provided props first so event handlers are never overridden */}
+        {
+          ...divProps /* expand user provided props first so event handlers are never overridden */
+        }
         onClick={this.composeHandlers(this.onClick)}
         onDragStart={this.composeHandlers(this.onDragStart)}
         onDragEnter={this.composeHandlers(this.onDragEnter)}
@@ -408,7 +412,9 @@ class Dropzone extends React.Component {
       >
         {this.renderChildren(children, isDragActive, isDragAccept, isDragReject)}
         <input
-          {...inputProps /* expand user provided inputProps first so inputAttributes override them */}
+          {
+            ...inputProps /* expand user provided inputProps first so inputAttributes override them */
+          }
           {...inputAttributes}
         />
       </div>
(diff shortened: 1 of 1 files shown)


thaddeusreid:
Hmm, looks like TravisCI is failing on a lint error that the `"precommit": "lint-staged"` will correct even when I am intentionally committing the lint changes TravisCI wants to see.

okonet:
I’m traveling now but I’ll check what’s wrong here next week and get back to you.

thaddeusreid:
Strange. The eslint phase of the TravisCI build asks for one thing:
```
/home/travis/build/react-dropzone/react-dropzone/src/index.js
  401:10  error  Replace `⏎··········...divProps·/*·expand·user·provided·props·first·so·event·handlers·are·never·overridden·*/⏎········` with `...divProps·/*·expand·user·provided·props·first·so·event·handlers·are·never·overridden·*/`        prettier/prettier
  415:12  error  Replace `⏎············...inputProps·/*·expand·user·provided·inputProps·first·so·inputAttributes·override·them·*/⏎··········` with `...inputProps·/*·expand·user·provided·inputProps·first·so·inputAttributes·override·them·*/`  prettier/prettier
```

Yet when I manually change my local file to match what Travis is expecting, the `precommit` script in package.json applies an `eslint --fix` to undo my change. Seems like there must be a version mismatch between my local and TravisCI.

Here are my local versions of eslint and lint-staged:
```bash
$ ./node_modules/.bin/eslint --version
v4.16.0
$ ./node_modules/.bin/lint-staged --version
4.3.0
```

okonet:
Weird. Travis uses the versions from `package.json` so I doubt there is a version mismatch. Must be something else.

okonet:
Have you tried running `eslint --fix` locally?

thaddeusreid:
@okonet yes.

`eslint --fix`  it styles the code like this:
```javascript
{
  ...divProps /* expand user provided props first so event handlers are never overridden */
 }
```
while TravisCI wants this:
```javascript
{...divProps /* expand user provided props first so event handlers are never overridden */}
```
This is so dirty... I removed `"precommit": "lint-staged",` from my local package.json so that I could commit the lint style TravisCI wanted... and now it's passing.

rkmarks:
To be honest @thaddeusreid I did the same for my contribution. It might have something to do w/ the way lint-staged is setup. I'll look into that. nice work though! 👍

src/index.js
```diff
@@ -239,11 +239,13 @@ class Dropzone extends React.Component {
...
       setTimeout(() => {
-        // Returns an object as FileList
-        const { files } = this.fileInputEl
+        if (this.fileInputEl != null) {
+          // Returns an object as FileList
+          const { files } = this.fileInputEl
 
-        if (!files.length) {
-          this.isFileDialogActive = false
+          if (!files.length) {
```
yuri-sakharov: simpler to add `if (!files || !files.length)` add remove this `if (this.fileInputEl != null) {`

okonet: I don't know. The PR code looks more readable to me

yuri-sakharov: for me more readable this version 
```
if (this.isFileDialogActive) {
                setTimeout(() => {
                    const { files } = this.fileInputEl;

                    if (!files || !files.length) {
                    }
                    if (typeof onFileDialogCancel === 'function') {
                        onFileDialogCancel();
                    }
                }, 300);
            }
```

yuri-sakharov: no need one more if and what about case with `files undefined`

okonet:
@rkmarks @thaddeusreid I just checked out the latest master and tried to commit some changes. `eslint --fix` did format it correctly (the way Travis CI would expect) so I'm not sure what's going on.

Please provide me with more details on how did you format the code so it was passing local checks but not the CI. One thing that can cause this is not latest dependencies being used. Please make sure you run `yarn install` before running `eslint --fix .` and `lint-staged`.

thaddeusreid:
@okonet Ahh, that's the problem. I installed dependencies with `npm install`.

I can see the same results as TravisCI after I removing node_modules, `yarn install`, and `yarn run eslint:src`.

Just a habit of mine to use `npm` over `yarn`. Thanks for the merge 👍

PR #564, opened 2018-01-30
umd build fix

franckamayou:

- [ ] bugfix
- [ ] feature
- [ ] refactoring / style
- [x] build / chore
- [ ] documentation

- [x] Not relevant

- [x] Not relevant

**Summary**

This pr solves a problem for the umd build with React loaded from the <script> tag as a global library. When including React from the script tag, it is available to the global namespace as the `window.React` object. However, when specifying the external library in webpack `externals: { "react" : "react" }`, it will return an error `Uncaught Error: Cannot find module "react".` since React is not available as a module but instead as a global object. Using `externals: { "react" : "React" }` instead will actually make Webpack fail to build, since the `import React from "react";` statement will not match any module.

It appears a solution, based on the comments on the link below (by user jide), is to specify `externals: { "react" : "umd react" }`, which takes care of both cases.

**Other information**

This is a [link](https://github.com/webpack/webpack/issues/1275
) describing the issue with webpack, packaging with react as an external library and umd.
Another [link](https://stackoverflow.com/a/34254167) from stackoverflow with another potential solution.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -26,8 +26,18 @@ module.exports = env => {
       extensions: ['.js', '.json']
     },
     externals: {
-      react: 'react',
-      'prop-types': 'prop-types'
+      react: {
+        root: 'React',
+        commonjs2: 'react',
+        commonjs: 'react',
+        amd: 'react'
+      },
+      'prop-types': {
+        root: 'PropTypes',
+        commonjs2: 'prop-types',
+        commonjs: 'prop-types',
+        amd: 'prop-types'
+      }
     }
   }
 }


okonet:
I'm not sure I'm familiar with this hack. Is it somewhere in docs?

Also, what is the problem exactly it solves? Please update the PR description.

franckamayou:
Updated the PR description. Let me know if this is clear enough.

rkmarks:
It is an interesting fix, however I think we should stay inline w/ the webpack documentation which is that if you are going to use react as a script vs in your bundle we should be splitting our component externals the way the stack overflow answer provided
```
externals: {
  "react": {
    root: 'React',
    commonjs2: 'react',
    commonjs: 'react',
    amd: 'react'
  }
}
```
that way when webpack creates the bundle to should look for each of those and use the one accessible
@okonet what do you think?

okonet:
Yes, I agree.

okonet:
Can you please try following the official way and check if it solves it?

franckamayou:
Sure thing, I'll let update the PR then

franckamayou:
I've updated the PR, tested and verified that it works with my application.

PR #573, opened 2018-02-19
npm now saves deps by default

pravdomil:
see diff


rkmarks:
I know it does it by default as of `npm5` however since this repo is also usable in older npm versions I believe it is fine to leave the docs showing `npm install --save` for clarity. @okonet what is your opinion?

okonet:
I agree, nothing bad about having `--save` in the docs.

PR #591, opened 2018-04-03
[WIP] feat: Add initial typescript definition

### LINKED ISSUE #404: Move Typescript definitions from definitely-typed into this package

- [ ] I found a bug
- [X] I want to propose a feature

Must install `@types/react-dropzone` to your project to use with typings

Include a `typings.d.ts` file directly in this project

* Fewer dependencies
* Nice to keep definitions close to source code
* Projects using `react-dropzone` don't have to worry about definitions not matching source.
* Versionin number makes more sense. i.e. current version of `@types/react-dropzone` is `0.0.32` and hasn't been updated in 2 months.
* Easier for me to contribute to source then to definitely typed.
* Remembering to keep up changes

patrick91:
I've ended up rewriting the TypeScript definition a bit. But I still would like to give credit to all the people that have worked on the [original ones](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-dropzone). Shall I add a comment on the top of the definition or is there a better way?

This PR might need some improvements still, I'll try to have a look at it in the next few days.

Closes #404


package.json
```diff
@@ -116,8 +119,11 @@
     "sinon": "^3.2.1",
     "size-limit": "^0.11.0",
     "style-loader": "^0.18.2",
+    "tslint": "^5.9.1",
```
okonet: Please add tslint to lint-staged config

okonet:
That's amazing. I have a question: if we change the implementation of the library, how do we ensure that types are up-to-date? Can this be added to the `test` script so it runs automatically on CI. Also, please add `tslint` to `lint-staged` config.

patrick91:
I'm definitively going to add the TSC compilation to the test script, so we can make sure we have valid typescript files.

Making sure that types are up to date won't be easy, I don't think there a way to manually check that

rkmarks:
@patrick91 have you had a chance to review and make changes per @okonet req? Hoping to get this added for typscripting

okonet:
Can someone else finish this up please?

package.json
patrick91: this is done, I'll be testing the PR on a project and update it as soon as I find any issue

patrick91:
@rkmarks just done :)

okonet:
Deploy should work now. If not please file an issue

PR #595, opened 2018-04-16
Drop and Drop Folders - keeps package smaller

veerabio:
This pull request includes code extracted from html5-file-selector and edited to keep the package size small

- [ ] bugfix
- [x] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Not relevant

**Summary**

Supports dropping a folder.

Yes, The Drop event now does not set the acceptedfiles and rejectedfiles on state.
This will affect anyone using these by providing a function as a children of Dropzone.

**Other information**

### DIFF
--- a/src/index.spec.js
+++ b/src/index.spec.js
@@ -633,49 +633,43 @@ describe('Dropzone', () => {
       expect(child).toHaveProp('isDragReject', false)
     })
 
-    it('should add valid files to rejected files on a multple drop when multiple false', done => {
+    it('should add valid files to rejected files on a multple drop when multiple false', async () => {
       const dropzone = mount(<Dropzone accept="image/*" onDrop={dropSpy} multiple={false} />)
       dropzone.simulate('drop', { dataTransfer: { files: images } })
-      setTimeout(() => {
-        const rejected = dropSpy.firstCall.args[0]
-        expect(rejected.length).toEqual(1)
-        done()
-      }, 0)
+      await null
+      const rejected = dropSpy.firstCall.args[0]
+      expect(rejected.length).toEqual(1)
     })
 
-    it('should add invalid files to rejected when multiple is false', done => {
+    it('should add invalid files to rejected when multiple is false', async () => {
       const dropzone = mount(<Dropzone accept="image/*" onDrop={dropSpy} multiple={false} />)
       dropzone.simulate('drop', {
         dataTransfer: { files: images.concat(files) }
       })
-      setTimeout(() => {
-        const rejected = dropSpy.firstCall.args[1]
-        expect(rejected.length).toEqual(2)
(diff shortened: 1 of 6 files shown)


src/index.spec.js
```diff
@@ -621,37 +633,49 @@ describe('Dropzone', () => {
       expect(child).toHaveProp('isDragReject', false)
     })
 
-    it('should add valid files to rejected files on a multple drop when multiple false', () => {
+    it('should add valid files to rejected files on a multple drop when multiple false', done => {
       const dropzone = mount(<Dropzone accept="image/*" onDrop={dropSpy} multiple={false} />)
       dropzone.simulate('drop', { dataTransfer: { files: images } })
-      const rejected = dropSpy.firstCall.args[0]
-      expect(rejected.length).toEqual(1)
+      setTimeout(() => {
```
okonet: I’d prefer us using async await here instead of timeouts.

okonet: Or anything that’s fits better from https://facebook.github.io/jest/docs/en/asynchronous.html#promises

rkmarks: I agree with that @okonet. We should be looking to move any promises to async await where we can.

veerabio: @okonet @rkmarks I prefer async/await as well, but in this case I think there is nothing to 'await' on. I will explore how I can do this, please suggest if you can think of a way

src/index.spec.js
```diff
@@ -633,49 +633,43 @@ describe('Dropzone', () => {
...
-    it('should add valid files to rejected files on a multple drop when multiple false', done => {
+    it('should add valid files to rejected files on a multple drop when multiple false', async () => {
       const dropzone = mount(<Dropzone accept="image/*" onDrop={dropSpy} multiple={false} />)
       dropzone.simulate('drop', { dataTransfer: { files: images } })
-      setTimeout(() => {
-        const rejected = dropSpy.firstCall.args[0]
-        expect(rejected.length).toEqual(1)
-        done()
-      }, 0)
+      await null
```
veerabio: @okonet this was the way to make it 'await' thus allowing the stub to resolve before doing the assertions

okonet:
I can't find the associated issue with the appropriate discussion on this one. Can you please link it here?

veerabio:
are you asking about the https://github.com/react-dropzone/react-dropzone/pull/595/commits/ae6716f3da99ffd28c9a462663479f34a3eb3e3c . this was a bug i found while using react-dropzone with folders. you can drag text and drop in the dropzone in which case the entry is null which causes issues. I made the same pull request to https://github.com/quarklemotion/html5-file-selector/pull/3

PR #601, opened 2018-05-01
build: bundle umd with rollup

TrySound:

- [ ] bugfix
- [ ] feature
- [ ] refactoring / style
- [x] build / chore
- [ ] documentation

- [x] Not relevant

- [x] Not relevant

**Summary**

Less code to parse especially with source maps

  webpack without sourcemaps - 15.35K
  webpack with sourcemaps - 120.2K
  rollup with sourcemaps - 61.12K
  rollup without sourcemaps - 13.74K

No


okonet:
Since webpack remains a hard dependency I don’t think it’s a good thing to introduce new bundle tool to the tool chain. This will make project harder to maintain and contribute.

TrySound:
That's why I suggested my collaboration. I can support both tools. And it's not so hard to support them, specifically rollup.

TrySound:
I care about my dependencies and would like to optimise them as much as it possible for end developer (myself).

TrySound:
Also I tend to simplify toolchain so you won't have any problems with it.

okonet:
Let's not introduce rollup yet, please. If we going t switch to it, we should switch the whole build to it. Please be aware that it's using react-styleguidist for docs which uses webpack config.

TrySound:
What's wrong with using both? Sure webpack is a tool of styleguidist here. I won't touch it. Rollup is a tool for better bundles quality. Don't care about complexity now. I will handle it.

okonet:
> What's wrong with using both?

I already replied: This will make project harder to maintain and contribute.

> You should not care about complexity now. I will handle it.

I care about that, though. Since I'm not familiar with the tools and I don't think that the added complexity is justified in this particular case.

> Sure webpack is a tool of styleguidist here.

Webpack is just fine for bundles, too. We're talking about several KBs of uncompressed code. And I believe we can optimize webpack build even further. Most importantly, GZiped this lib is at ~5 KB and I believe that rollup won't improve this number.

TrySound:
Gzip is not enough. Why you don't care parsing cost?

TrySound:
This complexity is temporary. Also rollup config is dead simple since you have an example and will be even simpler after a couple of PRs.

TrySound:
Let's just keep this PR open for a while.

okonet:
Since latest version of styleguidist doesn't rely on webpack config anymore and can use it internally, I think we should consider merging this rather than trying to migrate to Webpack 4.

okonet:
@TrySound can you please add some more information on how Rollup makes the UMD build smaller into the PR description?

Issue #523, opened 2017-10-24
Add support to required input


rafacianci:
**Do you want to request a *feature* or report a *bug*?**

- [ ] I found a bug
- [X] I want to propose a feature

**What is the current behavior?**

**If the current behavior is a bug, please provide the steps to reproduce.**

1.
1.
1.

**What is the expected behavior?**

**If this is a feature request, what is motivation or use case for changing the behavior?**
I want that the field provide the required props.
To set on HTML input that the field is required. :)
**Please mention other relevant information such as the browser version, Operating System and react-dropzone version.**

rkmarks:
Closing issue as you can pass the required field on the inputProps

PR #606, opened 2018-05-10
Enable user editing of inputAttributes.style

marcofugaro:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [x] Not relevant

**Summary**
In #523 you said you can pass `required` to the inputProps to enable form validation. In reality this doesn't work because the input element needs to be visible to be validated, otherwise this error happens.

screen shot 2018-05-10 at 19 13 17

All because of this line:
https://github.com/react-dropzone/react-dropzone/blob/72abe1d900fecd7784e08010e2f60bcb790c956a/src/index.js#L377

This pr allows the user to edit the style of the input element using `inputProps.style`

No

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -374,7 +374,7 @@ class Dropzone extends React.Component {
       accept,
       disabled,
       type: 'file',
-      style: { display: 'none' },
+      style: { display: 'none', ...inputProps.style },
       multiple: supportMultiple && multiple,
       ref: this.setRefs,
       onChange: this.onDrop,


src/index.js
```diff
@@ -374,7 +374,10 @@ class Dropzone extends React.Component {
       accept,
       disabled,
       type: 'file',
-      style: { display: 'none' },
+      style: {
+        display: 'none',
```
okonet: Should it instead remove `display: none` and use a different hiding technique?

src/index.js
```diff
@@ -374,7 +374,10 @@ class Dropzone extends React.Component {
...
+        ...(inputProps && inputProps.style)
```
okonet: Do you need to check for existence of `style` here? AFAIK, you can use `...` with `undefined` value?

src/index.js
marcofugaro: The only technique of hiding I found that still pops up the required notifier is `opacity: 0.0001` and `pointer-events: none`. At least in chrome.

Then I set a 
```css
position: absolute;
top: 0;
left: 0;
bottom: 0;
right: 0;
```
to align the required warning to the dropzone.

What do you think?

src/index.js
marcofugaro: Yea, it works fine with undefined, I tested even in safari:

screen shot 2018-05-25 at 15 09 47

src/index.js
okonet: I believe @matuzo knows a better trick.

src/index.js
okonet: Can you update the code please?

marcofugaro: @okonet what do you mean? Update the code how?

okonet: Remove the `(inputProps &&` part since it's not needed. `...inputProps.style` should be enough if `inputProps` would be set to `{}` in `defaultProps`?

marcofugaro: oh sure thing, sorry it was not clear

src/index.js
okonet: Please check out https://github.com/h5bp/html5-boilerplate/blob/7e74f27208ef6e62cda92057dfb853b51d3649dd/src/css/main.css#L121-L140

marcofugaro: @okonet it shows the tooltip, but it doesn't work with the positioning of the tooltip, in some browsers the tooltip is down, other up (safari), this is an example on safari:
image

With my solution the appear fine because of the position: absolute.

okonet: AFAIK the solution I proposed is battle tested in all modern browsers. In which browsers did you tests yours?

marcofugaro: @okonet Chrome, safari and firefox. I have no easy access to edge but I'm sure it works just fine

okonet: @marcofugaro okay let's go with your for now and see if people will report issues :)

marcofugaro: Sure thing!

marcofugaro:
@okonet added the styles for the input, but I had some troubles with the `position: relative;` on the container.

I updated the tests inserting `position: relative;` inline, did I do it right? Could it be more elaborate? Thanks

PR #619, opened 2018-06-07
Do not call any of the onDrag*/onDrop handlers if the user is not dragging files

rolandjitsu:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Not relevant

**Summary**

As this library only covers drag 'n drop of files, accidentally dragging text or html elements should not trigger any of the handlers attached to the dropzone (e.g. `onDragEnter`, `onDrop`, etc.).

A similar issue is also reported in https://github.com/react-dropzone/react-dropzone/issues/599.

Yes

**Other information**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,7 +10,8 @@ import {
   fileMatchSize,
   onDocumentDragOver,
   getDataTransferItems as defaultGetDataTransferItem,
-  isIeOrEdge
+  isIeOrEdge,
+  hasFiles
 } from './utils'
 import styles from './utils/styles'
 
@@ -84,9 +85,11 @@ class Dropzone extends React.Component {
   }
 
   onDragStart(evt) {
-    if (this.props.onDragStart) {
-      this.props.onDragStart.call(this, evt)
-    }
+    Promise.resolve(this.props.getDataTransferItems(evt)).then(draggedFiles => {
+      if (hasFiles(draggedFiles) && this.props.onDragStart) {
+        this.props.onDragStart.call(this, evt)
+      }
+    })
   }
 
   onDragEnter(evt) {
@@ -100,14 +103,17 @@ class Dropzone extends React.Component {
     evt.persist()
 
     Promise.resolve(this.props.getDataTransferItems(evt)).then(draggedFiles => {
-      this.setState({
-        isDragActive: true, // Do not rely on files for the drag state. It doesn't work in Safari.
-        draggedFiles
-      })
+      if (hasFiles(draggedFiles)) {
+        this.setState({
+          isDragActive: true, // Do not rely on files for the drag state. It doesn't work in Safari.
+          draggedFiles
+        })
+
+        if (this.props.onDragEnter) {
+          this.props.onDragEnter.call(this, evt)
+        }
+      }
     })
-    if (this.props.onDragEnter) {
-      this.props.onDragEnter.call(this, evt)
-    }
   }
 
   onDragOver(evt) {
@@ -123,9 +129,12 @@ class Dropzone extends React.Component {
       // continue regardless of error
     }
 
-    if (this.props.onDragOver) {
-      this.props.onDragOver.call(this, evt)
-    }
+    Promise.resolve(this.props.getDataTransferItems(evt)).then(draggedFiles => {
+      if (hasFiles(draggedFiles) && this.props.onDragOver) {
+        this.props.onDragOver.call(this, evt)
+      }
+    })
+
     return false
   }
 
@@ -144,9 +153,11 @@ class Dropzone extends React.Component {
       draggedFiles: []
     })
 
-    if (this.props.onDragLeave) {
-      this.props.onDragLeave.call(this, evt)
-    }
+    Promise.resolve(this.props.getDataTransferItems(evt)).then(draggedFiles => {
+      if (hasFiles(draggedFiles) && this.props.onDragLeave) {
+        this.props.onDragLeave.call(this, evt)
+      }
+    })
   }
 
   onDrop(evt) {
@@ -210,7 +221,7 @@ class Dropzone extends React.Component {
         rejectedFiles.push(...acceptedFiles.splice(0))
       }
 
-      if (onDrop) {
+      if (hasFiles(fileList) && onDrop) {
         onDrop.call(this, acceptedFiles, rejectedFiles, evt)
       }
 
--- a/src/index.spec.js
+++ b/src/index.spec.js
@@ -15,7 +15,18 @@ const flushPromises = wrapper =>
 const Dropzone = require('./index')
 const DummyChildComponent = () => null
 
+const createFile = (name, size, type) => {
+  const file = new File([], name, { type })
+  Object.defineProperty(file, 'size', {
+    get() {
+      return size
+    }
+  })
+  return file
+}
+
 let files
+let nonFileItems
 let images
 
 const rejectColor = 'red'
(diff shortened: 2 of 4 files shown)


okonet:
I believe this should be done in the user land so you could decide how to react on such drags.

rolandjitsu:
Well, from the description (*Simple HTML5-compliant drag'n'drop zone **for files** ...*) I assumed this lib handles file drag 'n' drop, so it made sense to avoid a behaviour where the user can select a piece of text and drag 'n' drop it.

src/index.js
```diff
@@ -81,6 +81,14 @@ class Dropzone extends React.Component {
   }
 
   onDragStart(evt) {
+    const items = getDataTransferItems(evt)
+
+    // Unless there are files being dragged,
+    // do not call the onDragStart.
+    if (!Array.isArray(items) || !items.length) {
```
okonet: Please extract the check to a function or method and use it to make the code more DRY

src/index.js
```diff
@@ -89,14 +97,22 @@ class Dropzone extends React.Component {
   onDragEnter(evt) {
     evt.preventDefault()
 
+    const items = getDataTransferItems(evt)
```
okonet: `getDataTransferItems` is `async` now so I'm not sure it's going to work like this. Are tests passing?

src/index.js
```diff
@@ -183,8 +216,11 @@ class Dropzone extends React.Component {
       rejectedFiles.push(...acceptedFiles.splice(1))
     }
 
-    if (onDrop) {
-      onDrop.call(this, acceptedFiles, rejectedFiles, evt)
+    // If no files, do not call onDrop.
+    if (Array.isArray(fileList) && fileList.length) {
+      if (onDrop) {
+        onDrop.call(this, acceptedFiles, rejectedFiles, evt)
```
okonet: I'm wondering if we should still allow firing callbacks via a flag?

src/index.spec.js
```diff
@@ -7,6 +7,7 @@ const Dropzone = require(process.env.JEST_TARGET ? process.env.JEST_TARGET : './
 const DummyChildComponent = () => null
 
 let files
+let strItems
```
okonet: Please rename to something more verbose like `nonFileItems`

src/utils/index.js
```diff
@@ -14,7 +14,20 @@ export function getDataTransferItems(event) {
...
       // During the drag even the dataTransfer.files is null
       // but Chrome implements some drag store, which is accesible via dataTransfer.items
-      dataTransferItemsList = dt.items
+      // NOTE: Allow only files and retun the items as a list of File,
+      // see https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem for details
+
+      const items = []
+
+      // eslint-disable-next-line no-plusplus
+      for (let i = 0; i < dt.items.length; i++) {
```
okonet: I don't like imperative `for` loops TBH. Let's use `Array.prototype.filter` here

src/utils/index.spec.js
```diff
@@ -18,6 +18,30 @@ const files = [
   }
 ]
 
+const strItems = [
```
okonet: Same as before

src/utils/index.spec.js
```diff
@@ -64,7 +102,7 @@ describe('getDataTransferItems', () => {
     expect(res).toHaveLength(3)
   })
 
-  it('should prioritize dataTransfer.files over .files', () => {
+  it('should prioritize dataTransfer.files over .items', () => {
```
okonet: 👍

src/index.js
rolandjitsu: Hmm, I'm just trying to figure that out, but the only possibility is to wait for the promise to resolve before doing anything in these `onDrag*` callbacks. Is there any reason why this is now async?

src/index.js
```diff
@@ -120,9 +128,14 @@ class Dropzone extends React.Component {
       // continue regardless of error
     }
 
-    if (this.props.onDragOver) {
-      this.props.onDragOver.call(this, evt)
-    }
+    Promise.resolve(this.props.getDataTransferItems(evt)).then(draggedFiles => {
+      if (hasFiles(draggedFiles)) {
```
okonet: Combine 2 `if`s?

src/index.js
```diff
@@ -141,9 +154,13 @@ class Dropzone extends React.Component {
       draggedFiles: []
     })
 
-    if (this.props.onDragLeave) {
-      this.props.onDragLeave.call(this, evt)
-    }
+    Promise.resolve(this.props.getDataTransferItems(evt)).then(draggedFiles => {
+      if (hasFiles(draggedFiles)) {
```
okonet: Same here

src/index.js
```diff
@@ -204,8 +221,10 @@ class Dropzone extends React.Component {
         rejectedFiles.push(...acceptedFiles.splice(0))
       }
 
-      if (onDrop) {
-        onDrop.call(this, acceptedFiles, rejectedFiles, evt)
+      if (hasFiles(fileList)) {
```
okonet: Same here?

src/utils/index.js
```diff
@@ -15,7 +15,9 @@ export function getDataTransferItems(event) {
     } else if (dt.items && dt.items.length) {
       // During the drag even the dataTransfer.files is null
       // but Chrome implements some drag store, which is accesible via dataTransfer.items
-      dataTransferItemsList = dt.items
+      // Allow only files and retun the items as a list of File,
+      // see https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem for details
+      dataTransferItemsList = Array.prototype.filter.call(dt.items, file => file.kind === 'file')
```
okonet: So this change made me wondering if `hasFiles` is going to break folder drop integration.

rolandjitsu: Folder drop is not something that is in the official [spec](https://www.w3.org/TR/2011/WD-html5-20110525/dnd.html) as far as I know.

So, in my opinion, supporting this could be a risk if it will never be supported natively.

rolandjitsu:
Let me know if there's something else I need to do to get this merged.

src/utils/index.js
okonet: Folder drop is supported over plugins. See https://github.com/react-dropzone/react-dropzone/pull/616

okonet: To stay on the safe side, I'd propose to leave this part of code as is and do the check in the `hasFiles` since it's the only place where it's really relevant. Ideally, if you could help building an example with folders support, so we can test against it in the browser, would be even better. Let me know if you up to it and/or need any help with that. Thoughts?

okonet:
@rolandjitsu also I've thought and I think it's a breaking change after all. We'll change the behavior of how callback are being called internally so it _might_ break some implementations that rely on previous one. There is nothing for you to do about it. I'll release it as a breaking change when it's done.

src/utils/index.js
rolandjitsu: Fair enough. I'll try to make these changes and add an example within the next few days.

okonet: The documentation PR #662 has been merged.

rolandjitsu: Great. I moved the filtering logic in the `hasFiles` fn, but I also made sure that the default `getDataTransferItems` only returns `File` objects (by using [getAsFile](https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/getAsFile) on the item).

Should also work with folders as the fn in the https://github.com/react-dropzone/react-dropzone/pull/662 example will only return `File` objects which we expect to get.

jare-talend:
I try the latest version and the onDragEnter isn't more triggered when i drag a file from my file system.
We can reproduce this on the demo page.
react-dropzone

okonet:
Looks like regression to me :( please investigate and create a PR if possible.

rolandjitsu:
I'm seeing a couple of syntax errors as well in the console:
```text
Error: Minified React error #152; visit https://reactjs.org/docs/error-decoder.html?invariant=152&args[]=StateHolder for the full message or use the non-minified dev environment for full errors and additional helpful warnings. 
...
SyntaxError: await is only valid in async function
```

I wonder if it's related.

rolandjitsu:
I think I found the issue, I'll make a PR soon with a fix.

PR #620, opened 2018-06-13
Primitive Dropzone with render props

donysukardi:

- [ ] bugfix
- [x] feature
- [x] refactoring / style
- [x] build / chore
- [x] documentation

- [ ] Not relevant

- [ ] Not relevant

**Summary**
Picking up removal of style management done by @okonet's. Refactored Dropzone to use render props, giving flexibility to users to implement their own styles in userland.

Yes it does. It changes the render signature of Dropzone.

**Other information**
Merging this to https://github.com/react-dropzone/react-dropzone/pull/468 for further review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,21 +37,23 @@ class MyDropzone extends React.Component {
   render() {
     return (
       <Dropzone onDrop={this.onDrop}>
-        {
-          ({ isDragActive }) => {
-            return (
-              <div className={cx('dropzone', {
+        {({ getRootProps, getInputProps, isDragActive }) => {
+          return (
+            <div
+              {...getRootProps()}
+              className={cx('dropzone', {
                 'dropzone--isActive': isDragActive
-              })}>
+              })}
+            >
+              <input {...getInputProps()} />
               {
                 isDragActive ?
                   <p>Drop files here...</p> :
                   <p>Try dropping some files here, or click to select files to upload.</p>
               }
-              </div>
-            )
-          }
-        }
+            </div>
+          )
+        }}
       </Dropzone>
     );
   }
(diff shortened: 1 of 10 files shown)


okonet:
Have you seen https://github.com/react-dropzone/react-dropzone/pull/468?

donysukardi:
Yes. In fact this builds on top of that. The bare minimum way to render a Dropzone component would be

```jsx
<Dropzone onDrop={this.onDrop}>
  {({ getRootProps, getInputProps }) => (
    <div {...getRootProps()}>
      <input {...getInputProps()} />
    </div>
  )}
</Dropzone>
```

This leaves the rendering responsibility to users and Dropzone just provides prop-getters and states - https://github.com/donysukardi/react-dropzone/tree/feature/primitive-dropzone#render-prop-function.

okonet:
I like the idea a lot. One question which I'm not sure is if we should expose the rendering of `input` to the user land since it's crucial for function. My ideas was to render it automatically inside the root element. How much do we want users to be able to customize it?

donysukardi:
I suppose we could do the following, since the input does not necessarily have to be inside the rendered component.

```javascript
render() {
    // ...
    return (
      <Fragment>
        <input {...inputProps} />
        {renderFn(/* all render props */)}
      </Fragment>
    )
}

**Comment by okonet** [comment]
That's what my PR is doing now.

**Comment by donysukardi** [comment]
We could either render the input for them,

```javascript
render() {
  // ...
  return (
    <Fragment>
      <input {...inputProps} />
      {renderFn(/* renderProps */)}
    </Fragment>
  )
}
```

or pass `input` to the render function so that user can render that somewhere inside the root component

```javascript
render() {
  // ...
  return (
    renderFn({
      input, // this is the input node with all the necessary props
      /* other renderProps */
    })
  )
}
```

okonet:
My question is more on a API part: do we need (and why) to allow rendering input anywhere outside of root element?

okonet:
I disagree. It makes sense if the implementation can change. In case of dropzone a barely see a case where it’s required to put input element somewhere else outside of root Dom Node at the moment.

vladp:
if Input element is exposed as @donysukardi  suggests, would that help in integrating this component with react-native , (and other JS-based setups where there is no notion of a root browser window))

donysukardi:
@okonet but having the flexibility to define where to put the input is important as it won't cause any conflict in styling as we'll be leaving it totally to the users. How I see it is that users will create their own Upload component using react-dropzone *once* and use it in their applications.

README.md
```diff
@@ -37,57 +37,90 @@ class MyDropzone extends React.Component {
...
+below:
 
-// Call HoC with your component and options as arguments
-export default makeDropzone(MyDropzone, {
-  onDrop: (acceptedFiles, rejectedFiles) => {
-    // Do something with files
-    // i.e. pre-process, upload etc.
-  } 
-})
+### Prop Getters
```
okonet: This is going to be generated by styleguidist. Let's remove it

README.md
```diff
@@ -37,57 +37,90 @@ class MyDropzone extends React.Component {
...
+| property                               | type      | description                                             |
+| -------------------------------------- | --------- | --------------------------------------------------------|
+| `isDragActive`                         | `boolean` | Flag indicating whether an active drag is in progress   |
+| `isDragAccept`                         | `boolean` | Flag indicating whether the files dragged are accepted  |
+| `isDragReject`                         | `boolean` | Flag indicating whether the files dragged are rejected  |
+| `draggedFiles`                         | `array`   | List of files in active drag                            |
+| `acceptedFiles`                        | `array`   | List of accepted files in active drag                   |
+| `rejectedFiles`                        | `array`   | List of rejected files in active drag                   |
+
+### Custom refKey
```
okonet: This should move to examples

examples/Basic/Readme.md
```diff
@@ -4,11 +4,12 @@ This example renders a list of the dropped files.
 
 ```
 class Basic extends React.Component {
-  constructor() {
-    super()
+  constructor(props) {
```
okonet: Those `props` aren't used and thus not needed there.

PR #623, opened 2018-06-26
Call onChange from inputProps

tobilen:

- [X] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [X] Not relevant

**Summary**

When an `onChange`key is provided via `inputProps` it will  be overridden by the `onDrop`function (see https://github.com/react-dropzone/react-dropzone/blob/d12f9f615f2d908a1c8d6e268701026b4fc7917c/src/index.js#L385). This PR adds a behavior similiar to the inputProp click handler for the change event to make sure the callback doesn't get swallowed up.


Nodman:
Hi, thanks for submitting this PR!
I am a bit confused about this because you can access `onChange` event via `onDrop` prop, no need to pass `onChange` into `inputProps`.

Could you please explain more?

tobilen:
sure! the `onDrop` callback does not provide an event as argument, only an array of selected files. i want to get the actual `FileList` object that's provided as part of the event, because i can use it in turn to set the `files`/`dataTransfer` attribute on other inputs (see https://stackoverflow.com/a/47522812).

i also like all of our interactable components to have the same interface (i.e. onChange prop with event as argument).

I might create another PR in the future to add an `onChange` prop to react-dropzone itself, which always provides a change event as first argument, regardless of whether files were selected via drop or file dialog. But i figured the change proposed in this PR is a good first step.

But without looking at my specific use case, the actual issue i see in the current implementation is that the point of an escape hatch like `inputProps` is to give direct access to the underlying element. Currently any change event is being overwritten. That's why i checked `bugfix` and not `feature`.

okonet:
As @Nodman mentioned, this is implemented. Probably the documentation can be improved. Please feel free to submit a documentation PR.

PR #632, opened 2018-07-19
Basic example - spread out rest of the files as well

raunofreiberg:
Isn't the intended purpose to allow uploading multiple files? With the current example, the array of files is always rewritten with the file that was recently uploaded.


rkmarks:
Actually the basic example should be just showing how the drag and drop feature w/ react works. In reality this should be it's own example on how to handle multiple files. Can you please update this PR to show multiple file uploads on a new example?

okonet:
I agree. It's called basic for a reason.

PR #641, opened 2018-08-17
fix: Reject multiple file drop if single mode

### LINKED ISSUE #458: Reject all if multiple prop is false and multiple files are dropped

Schmitze333:
* remove obsolete test
  * change test to verify that a single invalid file is rejected
  * move all dropped files to rejected and clear accepted

  closes #458

- [ ] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [ ] Not relevant

**Summary**

**Other information**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -196,10 +196,10 @@ class Dropzone extends React.Component {
         }
       })
 
-      if (!multiple) {
+      if (!multiple && acceptedFiles.length > 1) {
         // if not in multi mode add any extra accepted files to rejected.
         // This will allow end users to easily ignore a multi file drop in "single" mode.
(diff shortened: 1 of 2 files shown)


okonet:
Semantic release release a patch instead of the major version. I’ll need to resolve this from my computer.

PR #643, opened 2018-08-18
Chore upgrade deps

okonet:

- [ ] bugfix
- [ ] feature
- [x] refactoring / style
- [x] build / chore
- [ ] documentation

- [x] Not relevant

- [x] Not relevant

**Summary**

Upgrade all the things and fix linting errors

No

**Other information**


package.json
```diff
@@ -75,40 +73,45 @@
...
+    "@commitlint/cli": "^7.0.0",
+    "@commitlint/config-angular": "^7.0.1",
+    "@commitlint/prompt": "^7.0.0",
+    "@commitlint/prompt-cli": "^7.0.0",
     "babel-cli": "^6.9.0",
-    "babel-core": "^6.9.1",
-    "babel-eslint": "^7.1.1",
-    "babel-jest": "^21.0.0",
+    "babel-core": "^6.26.3",
+    "babel-eslint": "*",
```
TrySound: Please, prefer normal caret version.

package.json
```diff
@@ -75,38 +73,43 @@
...
     "imagemin-pngquant": "^5.0.0",
-    "jest": "^21.0.1",
+    "jest": "^23.5.0",
     "jest-enzyme": "^3.8.2",
-    "lint-staged": "^4.1.0",
-    "markdownlint-cli": "^0.3.1",
-    "prettier": "^1.6.1",
+    "lint-staged": "^7.2.2",
+    "markdownlint-cli": "^0.13.0",
+    "prettier": "*",
```
TrySound: Same here. It's better to know which version you use. `yarn upgrade-interactive --latest` will let you upgrade easier.

src/index.js
```diff
@@ -415,12 +417,17 @@ class Dropzone extends React.Component {
       getDataTransferItems,
       ...divProps
     } = props
+    /* eslint-enable no-unused-vars */
 
+    /* eslint-disable jsx-a11y/no-static-element-interactions */
+    /* eslint-disable jsx-a11y/click-events-have-key-events */
```
TrySound: Maybe it's easier to disable these rules at all?

.eslintrc
```diff
@@ -1,31 +1,18 @@
...
-  "extends": [
-    "okonet",
-    "prettier"
-  ],
-  "plugins": [
-    "prettier"
-  ],
+  "extends": ["okonet/node"],
+  "parserOptions": {
+    "sourceType": "module"
```
okonet: The only reason is the rollup config. @TrySound is it worth using imports?

src/index.js
okonet: They will go away in #468

package.json
okonet: This is added by `install-peerdeps` so I don't want to lock it

.eslintrc
TrySound: But source code is also esm. What's the problem?

package.json
okonet: Same here

PR #655, opened 2018-08-28
Fixes calling the provided onDrop-function with no files

LangdalP:
We had an issue where the onDrop method (provided as a prop) was called with empty lists for both accepted and rejected files.

This is related to this issue: https://github.com/facebook/react/issues/8793
This is a workaround, there might be a prettier way to fix this.


LangdalP:
I accidentally submitted this to the upstream repo. My plan is to temporarily use this fix until we can submit an issue or a better PR.

PR #657, opened 2018-09-03
fix: persist synthetic event

### LINKED ISSUE #646: Synthetic Events Warnings

- [x ] I found a bug

After updating my dependencies, I now get this warning when dragging and dropping a folder or file on to the dropzone component:

```
Warning: This synthetic event is reused for performance reasons. If you're seeing this, you're accessing the property `dataTransfer` on a released/nullified synthetic event. This is set to null. If you must keep the original synthetic event around, use event.persist(). See https://fb.me/react-event-pooling for more information.
```

Then, the drop handler callback has `null` on the event data transfer object

It used to work, it should still work, unsure why

Chrome 68, MacOS High Sierra, npm:

```json
  "dependencies": {
    "babel-polyfill": "^6.26.0",
    "material-ui": "^1.0.0-beta.46",
    "material-ui-icons": "^1.0.0-beta.36",
    "prop-types": "^15.6.2",
    "react": "^16.4.2",
    "react-dom": "^16.4.2",
    "react-dropzone": "^4.3.1",
    "react-redux": "^5.0.7",
    "react-router-dom": "^4.3.1",
    "redux": "^3.7.2",
    "redux-thunk": "^2.3.0",
    "sha1": "^1.1.1",
    "whatwg-fetch": "^2.0.4"
  },
  "devDependencies": {
    "babel-core": "^6.26.3",
    "babel-eslint": "^8.2.6",
    "babel-loader": "^7.1.5",
    "babel-plugin-transform-object-rest-spread": "^6.26.0",
    "babel-preset-env": "^1.7.0",
    "babel-preset-react": "^6.24.1",
    "eslint": "^4.19.1",
    "eslint-config-airbnb": "^16.1.0",
    "eslint-loader": "^2.1.0",
    "eslint-plugin-import": "^2.14.0",
    "eslint-plugin-jsx-a11y": "^6.1.1",
    "eslint-plugin-react": "^7.11.1",
    "uglifyjs-webpack-plugin": "^1.3.0",
    "webpack": "^4.16.5",
    "webpack-cli": "^2.1.5"
  }
```

I've tried calling `Persist` to no avail here:

```js
const mapDispatchToProps = dispatch => ({
  onDropFiles: (accepted, rejected, evt) => {
    evt.persist();
    dispatch(filesDropped(accepted, rejected, evt));
  },
});
```

I get an error here that I didn't before:

```javascript
export function filesDropped(accepted, rejected, evt) {
  const data = evt.dataTransfer.items;
```

The `dataTransfer` object is `null` and it bails when trying to access the non-existent `items` parameter. `filesDropped` is passed in as the `onDrop` property:

```jsx
const FolderDropzone = ({ classes, onDrop }) => (
  <Grid item xs={12}>
    <Paper className={classes.paper}>
      <div className="dropzone_area">
        <Dropzone onDrop={onDrop} className="dropzone" activeClassName="dropzone_active" acceptClassName="dropzone_accept" rejectClassName="dropzone_reject" disabledClassName="dropzone_disabled">
          <Typography>Try dragging and dropping a folder here,<br /> or click to select files to scan.</Typography>
          <p><small>Requires Google Chrome</small></p>
        </Dropzone>
      </div>
    </Paper>
  </Grid>
);
```

dan-lee:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant
- [x] Don't know how

- [x] Not relevant

**Summary**

Since v4.3 the event passed to the handler is not reusable anymore, because some work is done with it before. To prevent this `evt.persist()` is called.

Fixes #646 

no

**Other information**
I don't really know how to test this.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -97,6 +97,8 @@ class Dropzone extends React.Component {
       this.dragTargets.push(evt.target)
     }
 
+    evt.persist()
+
     Promise.resolve(this.props.getDataTransferItems(evt)).then(draggedFiles => {
       this.setState({
         isDragActive: true, // Do not rely on files for the drag state. It doesn't work in Safari.
@@ -173,6 +175,8 @@ class Dropzone extends React.Component {
       isDragActive: false,
       draggedFiles: []
     })
+
+    evt.persist()
 
     Promise.resolve(getDataTransferItems(evt)).then(fileList => {
       const acceptedFiles = []
--- a/src/index.js
+++ b/src/index.js
@@ -162,6 +162,9 @@ class Dropzone extends React.Component {
 
     // Stop default browser behavior
     evt.preventDefault()
+    
+    // Persist event for later usage
+    evt.persist()
 
     // Reset the counter along with the drag on a drop.
     this.dragTargets = []
@@ -175,8 +178,6 @@ class Dropzone extends React.Component {
       isDragActive: false,
       draggedFiles: []
     })
-
-    evt.persist()
 
     Promise.resolve(getDataTransferItems(evt)).then(fileList => {
       const acceptedFiles = []
--- a/src/index.js
+++ b/src/index.js
@@ -162,7 +162,7 @@ class Dropzone extends React.Component {
 
     // Stop default browser behavior
     evt.preventDefault()
-    
+
     // Persist event for later usage
     evt.persist()
 


tomjn:
I tried something similar to fix that issue, it didn't work but I think the problem was I didn't do it in all the places you did :)

tomjn:
On further review, is it not needed in `onDragStart`, `onDragOver`, `onDragLeave`, and `onClick`? Also I would push the persist call further up, ideally with the preventDefault and stop propagation calls

dan-lee:
> On further review, is it not needed in onDragStart, onDragOver, onDragLeave, and onClick

Oh so the only place should be `onDrop` then, I guess? I just adjusted all the places where `getDataTransferItems` 'manipulated' the event.

> Also I would push the persist call further up, ideally with the preventDefault and stop propagation calls

Fair point! Will change it

okonet:
Can we add tests for this?

dan-lee:
@okonet I really would love to! But this behaviour doesn't seem to be reflected in the unit tests. The event is *always* populated with `dataTransfer`, which in practice it shouldn't. I tried a lot of things, but couldn't find a way to break it.

okonet:
Hmm, I guess we can merge this and add tests later then.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #662, opened 2018-09-08
docs: Add en example of folders drag'n'drop

### LINKED ISSUE #661: Add an example for folders drag'n'drop
As of #https://github.com/react-dropzone/react-dropzone/issues/609 and it's implementation in https://github.com/react-dropzone/react-dropzone/pull/616 we now officially support plugins that should allow integrations with plugins like https://github.com/quarklemotion/html5-file-selector a breeze. It would be great to have an example of such integration in the examples.

d-ivashchuk:

- [x] documentation

- [x] Not relevant

- [ ] Not relevant

**Summary**

**Other information**

Closes #661

### DIFF
--- a/examples/Folders/Readme.md
+++ b/examples/Folders/Readme.md
@@ -1,0 +1,49 @@
+Dropzone that accepts folders as drag-and-drop.
+
+```
+const {getDroppedOrSelectedFiles} = require('html5-file-selector')
+
+class FolderDropzone extends React.Component {
+  constructor() {
+    super()
+    this.state = { files: [] }
+  }
+
+  onDrop(files) {
+    this.setState({
+      files
+    })
+  }
+
+  render() {
+    return (
+      <section>
+        <div className="dropzone">
+          <Dropzone
+            getDataTransferItems={evt =>
+              getDroppedOrSelectedFiles(evt).then(list =>
+                list.map(({ fileObject, ...rest }) => fileObject)
+              )
+            }
+            onDrop={this.onDrop.bind(this)}
+          >
+            <p>Drop a folder with files here.</p>
+          </Dropzone>
+        </div>
+        <aside>
+          <h2>Dropped files</h2>
+          <ul>
+            {this.state.files.map(f => (
+              <li key={f.name}>
+                {f.name} - {f.size} bytes
+              </li>
+            ))}
+          </ul>
+        </aside>
+      </section>
+    )
+  }
+}
+
+<FolderDropzone />
+```
--- a/examples/Folders/Readme.md
+++ b/examples/Folders/Readme.md
@@ -1,7 +1,7 @@
-Dropzone that accepts folders as drag-and-drop.
+Dropzone that accepts folders as drag-and-drop. Supports multiple folders and subfolders.
 
-```
-const {getDroppedOrSelectedFiles} = require('html5-file-selector')
+```jsx harmony
+const { getDroppedOrSelectedFiles } = require('html5-file-selector')
 
 class FolderDropzone extends React.Component {
   constructor() {
@@ -20,22 +20,18 @@ class FolderDropzone extends React.Component {
       <section>
         <div className="dropzone">
           <Dropzone
-            getDataTransferItems={evt =>
-              getDroppedOrSelectedFiles(evt).then(list =>
-                list.map(({ fileObject, ...rest }) => fileObject)
-              )
-            }
+            getDataTransferItems={evt => getDroppedOrSelectedFiles(evt).then(files => files)}
             onDrop={this.onDrop.bind(this)}
           >
             <p>Drop a folder with files here.</p>
           </Dropzone>
         </div>
         <aside>
-          <h2>Dropped files</h2>
+          <h2>Dropped files and folders</h2>
           <ul>
             {this.state.files.map(f => (
               <li key={f.name}>
-                {f.name} - {f.size} bytes
+                {f.fullPath} - {f.size} bytes
               </li>
             ))}
           </ul>
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
     "eslint-plugin-react": "7.x",
     "husky": "^0.14.3",
     "imagemin-cli": "^3.0.0",
-    "imagemin-pngquant": "^5.0.0",
+    "imagemin-pngquant": "^6.0.0",
     "jest": "^23.5.0",
     "jest-enzyme": "^6.0.3",
     "lint-staged": "^7.2.2",
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
     "logo": "cd logo && sketchtool export artboards logo.sketch",
     "imagemin": "imagemin --out-dir=logo --plugin=pngquant --plugin=svgo",
     "size": "size-limit",
-    "size:why": "size-limit --why"
+    "size:why": "size-limit --why",
+    "typescript": "tsc --project ./typings/tests"
   },
   "size-limit": [
     {
@@ -35,6 +36,9 @@
     "*.js": [
       "eslint --fix",
       "git add"
+    ],
+    "*.ts": [
+      "tslint"
     ],
     "*.{svg,png}": [
       "imagemin",
@@ -81,6 +85,8 @@
     "@commitlint/config-angular": "^7.0.1",
     "@commitlint/prompt": "^7.0.0",
     "@commitlint/prompt-cli": "^7.0.0",
+    "@types/react": "^16.3.2",
+    "@types/react-dom": "^16.0.4",
     "babel-cli": "^6.9.0",
     "babel-core": "^6.26.3",
     "babel-eslint": "*",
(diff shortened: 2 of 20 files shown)


rkmarks:
@d-ivashchuk can you update from master please :)

package.json
```diff
@@ -78,6 +78,7 @@
   },
   "dependencies": {
     "attr-accept": "^1.1.3",
+    "html5-file-selector": "^2.1.0",
```
okonet: This needs to go into `devDependencies` since it should not become part of the distribution package.

styleguide.config.js
```diff
@@ -28,6 +28,10 @@ module.exports = {
           name: 'Basic example',
           content: 'examples/Basic/Readme.md'
         },
+        {
+          name: 'Dragging folders example',
```
okonet: Can this be pushed below, please? It should be the last example since it's the most advanced one.

examples/Folders/Readme.md
```diff
@@ -0,0 +1,49 @@
+Dropzone that accepts folders as drag-and-drop.
+
+```
```
okonet: This is missing language. Please add `jsx harmony` here. This will enable syntax highlighting on GitHub and code editors.

examples/Folders/Readme.md
```diff
@@ -0,0 +1,49 @@
...
+const {getDroppedOrSelectedFiles} = require('html5-file-selector')
```
okonet: It looks like this code didn't run through prettier before committing.  AFAIK it should have white spaces inside `{}`

examples/Folders/Readme.md
```diff
@@ -0,0 +1,49 @@
...
+          >
+            <p>Drop a folder with files here.</p>
+          </Dropzone>
+        </div>
+        <aside>
+          <h2>Dropped files</h2>
+          <ul>
+            {this.state.files.map(f => (
+              <li key={f.name}>
+                {f.name} - {f.size} bytes
```
okonet: We should display the folder structure here so the example is clear about what it implements. Does the plugin we're using supports that? If not, we should at least show full paths here.

examples/Folders/Readme.md
```diff
@@ -0,0 +1,49 @@
...
+    })
+  }
+
+  render() {
+    return (
+      <section>
+        <div className="dropzone">
+          <Dropzone
+            getDataTransferItems={evt =>
+              getDroppedOrSelectedFiles(evt).then(list =>
```
okonet: Hmm, why is the `map` needed? Is this something coming from the plugin?

okonet:
> update from master

@d-ivashchuk you already have done what @rkmarks was asking by clicking on "Update branch"

d-ivashchuk:
@okonet ok, will work on that in the evening!

examples/Folders/Readme.md
d-ivashchuk: @okonet  I am not sure about this one. Shall I just add `// language=JSX Harmony` at the top of this file? It is the only thing I found for setting language explicitly, is it right?

rkmarks: no it would just need to be added like so ` ```jsx harmony`

examples/Folders/Readme.md
d-ivashchuk: @okonet No, you are right it's redundant

examples/Folders/Readme.md
d-ivashchuk: @rkmarks Thanks!

examples/Folders/Readme.md
d-ivashchuk: I couldn't find anything to display folder structure in the plug-in, so I will add full paths of dropped files for now so that it's clear what plug-in is responsible for.

d-ivashchuk:
I've tried to cover all the requested changes in commit `be2a4fc`. The one with an id of `ff887d2` is here just because I've messed up with git little bit :)

okonet:
> The one with an id of ff887d2 is here just because I've messed up with git little bit

Just wondering if you know about `git rebase` already?

examples/Folders/Readme.md
```diff
@@ -0,0 +1,45 @@
...
+      files
+    })
+  }
+
+  render() {
+    return (
+      <section>
+        <div className="dropzone">
+          <Dropzone
+            getDataTransferItems={evt => getDroppedOrSelectedFiles(evt).then(files => files)}
```
okonet: Since the `html5-file-selector` resolves with `files` already and this is what we expect according to https://github.com/react-dropzone/react-dropzone/pull/616/files#diff-1fdf421c05c1140f6d71444ea2b27638R97 I'm wondering if this is unnecessary to make it that explicit here.

This line can be replaced with 

```jsx
getDataTransferItems={getDroppedOrSelectedFiles}
```

On the other hand, this shows all possible hooking points for it.

I'm not sure so please express your opinion on this one. What would you prefer seeing in the example? A shorter version or this one?

examples/Folders/Readme.md
```diff
@@ -0,0 +1,45 @@
+Dropzone that accepts folders as drag-and-drop. Supports multiple folders and subfolders.
```
okonet: I tested the example and it works great. But the documentation still missing bits on the plugin architecture IMO. I would add more information about the architecture and emphasize the pluggable part. We would need to show function signature and an isolated example with promises IMO.

styleguide.config.js
```diff
@@ -43,6 +43,10 @@ module.exports = {
         {
           name: 'Full Screen Dropzone',
           content: 'examples/Fullscreen/Readme.md'
+        },
+        {
+          name: 'Dragging folders example',
```
okonet: Keeping in mind my first comment I'm wondering if this should become a more generic name. Like "Extending Dropzone"? Or "Custom `getDataTransferItems`"?

d-ivashchuk:
>> The one with an id of ff887d2 is here just because I've messed up with git little bit ...

@okonet kind of, but I’ve never worked with git repos that are dynamically changing before so I am still learning some thing. This one happened when my branch fell behind with the master and I needed to update my forked branch before pushing changes.

examples/Folders/Readme.md
d-ivashchuk: For me the explicit example is more about readability and understanding the code at once. It took me some time before I managed to understand how everything works, and my example reflects that I think. I’ve tried to make it self explanatory but of course if we explain the plugin architecture in docs, and how everything works under the hood there is no need of such explicitness.

styleguide.config.js
d-ivashchuk: Yes, I agree, one of those names would be more suitable. But at the same time I think people are looking for solving specific problem when they are reading docs so I would suggest making a section “Extending Dropzone” and introduce some subsections to it like the one with Dropzone that accepts folders. I deem it more logical while all the value of libraries lies within the simplicity of use in every particular use-case, maybe after we can also think of other examples and add more subsections to make the info flow more pleasant. What do you think?

examples/Folders/Readme.md
d-ivashchuk: Do you mean that we need to explain what the plugin does(how it makes folder processing possible) and how the returned value(list of files in the folder) is then used in Dropzone component utilizing promises? I may think not that deep but in the light of my comment regarding naming maybe we could overload this component users little bit less with how it works and emphasize more what it solves. But anyways I think both variants are possible and docs can also contain info on the architecture.
If you want me to add it I think I need a little explanation of what you mean by function signature and isolated example though

okonet: @d-ivashchuk no I mean the API part. Function signature etc. Your example is just one case of what's possible so we need to add this to the general docs (ideally this would live inside the component using JSDoc and styleguidist will extract this) but not to this example.

styleguide.config.js
okonet: I think `Extending Dropzone` as a separate page and then sub-sections with different examples could work best.

okonet:
@d-ivashchuk sorry I completely missed your replies. GitHub didn't notify me about them. Do you think you could finalize this PR this week so we can merge it. As a benefit you'd get closer to [Hacktoberfest](https://hacktoberfest.digitalocean.com) goal.

d-ivashchuk:
@okonet don't worry, I thought the same and anyway wanted to ping you on this tomorrow:) I will do my best to finish it

d-ivashchuk:
@okonet I am not sure I am doing the right thing, so have a look at the last changes please. I've rearranged the structure of `Extending Dropzone` section so it contains subsections and added a little bit of info to the `Plugin Architecture`. As far as I can see now, `getDataTransferItems` prop which is where we plug our functions already has function signature and is there in docs. Maybe I understood something wrong, if so, it would be great if you could elaborate a little bit.

okonet:
I think that’s perfect! I’m going to merge it. Thanks again for the help and pulling this off.

d-ivashchuk:
I am happy to help! If you have some other tasks that need some help, feel free to ping me!

PR #673, opened 2018-10-02
feat: improve TypeScript setup

rolandjitsu:
**What kind of change does this PR introduce?**

- [ ] bugfix
- [x] feature
- [ ] refactoring / style
- [x] build / chore
- [ ] documentation

**Did you add tests for your changes?**

- [x] Yes, my code is well tested
- [ ] Not relevant

**If relevant, did you update the documentation?**

- [ ] Yes, I've updated the documentation
- [x] Not relevant

**Summary**

This mainly adds support for using [createRef()](https://reactjs.org/docs/refs-and-the-dom.html#creating-refs) with the `{ref}` property on the `<Dropzone>` component.

Besides that:

- run `tsc` on the .tsx test files during tests (to ensure it still compiles if changes occur)
- add missing `tsconfig.json` file (otherwise the `tsc` command would fail)
- use default React export in .tsx files
- add missing type for `{getDataTransferItems}` prop
- update react typings
- update npm package lock

**Does this PR introduce a breaking change?**

No.

**Other information**

PR #678, opened 2018-10-03
fix: fix file filter PR regression

rolandjitsu:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Not relevant

**Summary**

Provide fix for https://github.com/react-dropzone/react-dropzone/pull/619#issuecomment-426573278.

No.

**Other information**
It also fixes some of the errors from the docs page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,13 +95,12 @@ See https://react-dropzone.netlify.com/#proptypes
 ### Testing
 
 *Important*: `react-dropzone` makes its drag'n'drop callbacks asynchronous to enable promise based getDataTransfer functions. In order to properly test this, you may want to utilize a helper function to run all promises like this:
-```js
+```js static
 const flushPromises = () => new Promise(resolve => setImmediate(resolve));
 ```
 
 Example with enzyme 3:
-```js
-
+```js static
 it('tests drag state', async () => {
   const flushPromises = () => new Promise(resolve => setImmediate(resolve));
   const DummyChildComponent = () => null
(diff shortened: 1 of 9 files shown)


rolandjitsu:
@okonet note that the folder drop example is not going to work because the plugin that it's used to illustrate this case does not return a list of native `File` objects.

The only viable option I see here is to extend the native `File` object and add the additional props, but this is something that the plugin author would need to do.

Alternatively, I could make another plugin which would do just that.

okonet:
That’s bummer. Does this mean the plugin we’re using becomes incompatible with this fix? What would the fix for it look like?

rolandjitsu:
Pretty much. We could add a specific check for the object type the plugin returns (see [Html5FileSelector.js#L71](https://github.com/quarklemotion/html5-file-selector/blob/master/src/Html5FileSelector.js#L71)), but I think that would be a mistake.

I think that plugins should comply with the type signature that this component requires for `getDataTransferItems()` (as illustrated in [#673](https://github.com/react-dropzone/react-dropzone/pull/673/files#diff-7359d6ad5f8dd67c439f94fc95f12d5eR57)). 

In any case, allowing plugins to return any kind of objects might cause some confusion for users as they would probably expect the callbacks (`onDrop()`, etc.) to be called with `File` (`FileWithPreview`) objects which they then can upload to a server (as an example).

So I guess we could either:

- make plugins comply with the same type signature for `getDataTransferItems()` and I could make another plugin which would do that (or make a PR for the current one)
- or expose `isFileList()` as a property (under some other name) which the plugin would also have to provide.

okonet:
I agree with the argument that users expect `File`s to be returned. Let's do the following:

1. Merge and release this PR
1. Make a PR for the current plugin for folders
1. Update the documentation for the function signature to make it clear
1. Add a test (not sure this is not covered yet)

PR #680, opened 2018-10-04
fix(edge): drag a link trigger drop event

jare-talend:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Not relevant

**Summary**

Drag a link with edge triggers a drop event
edge
 

**Other information**
When we drag a link on the dropzone, edge considers it like a file.
With edge, on dragenter, getDataTransferItems returns an items object because files isn't still available. so, the object has a string kind.
on drop event, the files property is available and like lit is prior, getDataTransferItems return it, then files has a File instance and isFileList detects it like a file.
We have to check on the items property if all the dragged files are like a file.


jare-talend:
#https://github.com/react-dropzone/react-dropzone/issues/681

rolandjitsu:
I think there might be a simpler fix for this. I don't have Edge to test this out, but I think that it should suffice to just call `isFileList(fileList)` at [index.js#L196](https://github.com/react-dropzone/react-dropzone/blob/master/src/index.js#L196) and return if it's `false`.

```js
const acceptedFiles = []
const rejectedFiles = []

if (!isFileList(fileList)) {
  return;
}

fileList.forEach(file => {
```

src/index.js
```diff
@@ -194,34 +195,39 @@ class Dropzone extends React.Component {
...
       const rejectedFiles = []
 
-      fileList.forEach(file => {
-        if (!disablePreview) {
-          try {
-            file.preview = window.URL.createObjectURL(file) // eslint-disable-line no-param-reassign
-          } catch (err) {
-            if (process.env.NODE_ENV !== 'production') {
-              console.error('Failed to generate preview for file', file, err) // eslint-disable-line no-console
+      if (!isFileDataTransferItems(evt)) {
```
rolandjitsu: Actually, all you need here is:
```js
if (
  !isFileList(fileList) ||
  (isIeOrEdge() && evt.dataTransfer && !isFileList(Array.from(evt.dataTransfer.items)))
) {
  return
}
```

Everything else in the PR can be removed. And you need to fix one of the preview test cases.

I'd also add some tests for this use case to ensure everything works as expected when IE/Edge is used.

I tested this in Edge 17/16 on SauceLabs.

src/index.js
```diff
@@ -194,6 +194,11 @@ class Dropzone extends React.Component {
       const acceptedFiles = []
       const rejectedFiles = []
 
+      if (isIeOrEdge() && evt.dataTransfer && !isFileList(Array.from(evt.dataTransfer.items))) {
+        rejectedFiles.push(...fileList)
```
rolandjitsu: I'm not so sure that we should reject in this case, because what you actually dropped are not files (technically Edge says they are, when in fact they are urls). I think we should only reject when the accepted mime type does not match with the files that are dropped.

src/index.spec.js
```diff
@@ -910,6 +910,22 @@ describe('Dropzone', () => {
       expect(onDropAccepted).toHaveBeenCalledWith(files.concat(images), expectedEvent)
       expect(onDropRejected).not.toHaveBeenCalled()
     })
+
+    it('should reject the no files (drag a with edge)', async () => {
```
rolandjitsu: I also think you need more tests as coverage has dropped.

src/index.js
rolandjitsu: You're also missing the first `!isFileList(fileList)`:
```js
if (
  !isFileList(fileList) ||
  (isIeOrEdge() && evt.dataTransfer && !isFileList(Array.from(evt.dataTransfer.items)))
) {
  return
}
```

rolandjitsu:
@jare-talend could you make sure that coverage doesn't drop? It's probably the new `if` that was added, but you can see the problem when running the tests locally.

src/index.spec.js
```diff
@@ -939,13 +956,10 @@ describe('Dropzone', () => {
 
     it('should not throw error when preview cannot be created', async () => {
```
rolandjitsu: We could rename this test case to `should not generate previews for non-File items`

src/index.spec.js
```diff
@@ -910,6 +910,23 @@ describe('Dropzone', () => {
       expect(onDropAccepted).toHaveBeenCalledWith(files.concat(images), expectedEvent)
       expect(onDropRejected).not.toHaveBeenCalled()
     })
+
+    it('should reject the non-files (drag an link a with edge)', async () => {
```
rolandjitsu: Rename this to `should not call onDrop* callbacks in Edge for non-File items`.

rolandjitsu:
And make sure to rebase your local branch from master as there has been some changes merged recently which conflict with yours.

rolandjitsu:
@jare-talend there are still some conflicts, could you update your branch using rebase so we can merge this?

rolandjitsu:
@jare-talend I'd advise [squashing](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) all your commits into a single commit (including [rebase from upstream](https://gist.github.com/ravibhure/a7e0918ff4937c9ea1c456698dcd58aa)), otherwise we cannot easily merge your PR as it seems to be missing some changes.

PR #682, opened 2018-10-05
fix: update folder drop example

rolandjitsu:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [x] documentation

- [ ] Not relevant

- [ ] Not relevant

**Summary**

Fixed the folder drop example (used a different lib) and added some tests, and added more docs on plugins.

No.

**Other information**

### DIFF
--- a/examples/Folders/Readme.md
+++ b/examples/Folders/Readme.md
@@ -1,7 +1,7 @@
 Dropzone that accepts folders as drag-and-drop. Supports multiple folders and subfolders.
 
 ```jsx harmony
-const { getDroppedOrSelectedFiles } = require('html5-file-selector')
+const { fromEvent } = require('file-selector')
 
 class FolderDropzone extends React.Component {
   constructor() {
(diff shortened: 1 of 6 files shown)


rolandjitsu:
@okonet this should cover the plugin issue and folder support.

rolandjitsu:
@okonet do you think we could get this PR and https://github.com/react-dropzone/react-dropzone/pull/673 (after I rebase it) merged?

okonet:
@rolandjitsu I've added you as a collaborator so you're free to merge this PR.

rolandjitsu:
@okonet to answer your questions:

1. In essence, it does achieve everything that the other lib did with some slight differences:
a. It's written in TS (makes it easier to maintain typings)
b. It returns native `File` objects patched with a `{path}` property
c. Provides es5 bundles (besides cjs, umd and es)
c. I chose to make a different lib because I thought it would be much faster to do so than make a PR with all the things + making releases
2. Do you mean replace the current default `getDataTransferItems()` with the one provided by the lib?
3. I am not against that, do you think it would be better to be in its own repo given that it's written in TS? And perhaps move it to this repo later on when/if this repo would also be written in TS (I think it would be a good idea and would help with maintaining typings)?

rolandjitsu:
@okonet thanks a bunch, I appreciate that.

okonet:
I’m open to TS in this repo if you’re up for it BTW.

okonet:
And to 3. I think it’s own repo is fine as long it is in this org

rolandjitsu:
Ok, I will move the lib to this org and make sure the auto publish from Travis still works. I'm open to work on migrating to TS, but later on when I have more time on my hands.

Issue #40, opened 2015-06-16
onDrop does not fire for same filename with different content


themuvarov:
Steps to reproduce:
1. Choose a file A from file system in dropzone
2. onDrop does fire, the file eventuanaly sent
3. Modify body of the file A
4. Choose the same file
5. onDrop does not fire since it mapped to onChange form's event

Could fix if by adding the following line to onClick function.

React.findDOMNode(this.refs.fileInput).value = null;

paramaggarwal:
We are using the HTML `onDrop` attribute on the `div` tag. If that's the default behaviour, then so be it. I don't see a reason to modify default behaviour here.

themuvarov:
it was eventually fixed here https://github.com/paramaggarwal/react-dropzone/commit/8bcf347f8f1b5dd6c8ad2498b978772e3ff351dc

mcmillhj-wta:
I am still experiencing this in 6.0.2 after https://github.com/react-dropzone/react-dropzone/commit/8bcf347f8f1b5dd6c8ad2498b978772e3ff351dc

PR #683, opened 2018-10-05
fix: pass event to user-supplied inputProps {onClick}

mcmillhj:

- [X] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [X] Not relevant

- [X] Not relevant

**Summary**

We encountered an issue that was described here, #40, that describes attempting to load the same file twice (the data has changed). The use-case where we encountered this was a user uploaded a file, but the rejected it due to formatting issues. The file input doesn't discard the target state so uploaded the correct file moments later does not trigger any handles such as `onDrop`, etc....

Summary of changes:

1. Update that event is passed to `onClick` prop handler inside of
`onInputElementClick`
2. Update `'should invoke inputProps onClick if provided'` testcase to
verify that function is now called with a MouseEvent object

No, this PR updates the application to match the documentation for `onClick`

**Other information**
- No docs were added or updated because it was a bugfix to make the code _match_ the current documentation 
- A single testcase was updated to verify that the a user supplied `onClick` handler now receives a `MouseEvent` object

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -263,7 +263,7 @@ class Dropzone extends React.Component {
   onInputElementClick(evt) {
     evt.stopPropagation()
     if (this.props.inputProps && this.props.inputProps.onClick) {
-      this.props.inputProps.onClick()
+      this.props.inputProps.onClick(evt)
     }
   }
 
--- a/src/index.spec.js
+++ b/src/index.spec.js
@@ -295,7 +295,7 @@ describe('Dropzone', () => {
       const component = mount(<Dropzone inputProps={{ onClick }} />)
 
       component.simulate('click')
-      expect(onClick).toHaveBeenCalled()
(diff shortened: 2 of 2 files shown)


src/index.js
```diff
@@ -263,7 +263,7 @@ class Dropzone extends React.Component {
   onInputElementClick(evt) {
     evt.stopPropagation()
     if (this.props.inputProps && this.props.inputProps.onClick) {
-      this.props.inputProps.onClick()
+      this.props.inputProps.onClick(evt)
```
rolandjitsu: There is a test case (`should invoke inputProps onClick if provided`) which checks the `onClick()` behaviour, could you update the test to reflect your changes (cb is called with evt)?

mcmillhj: definitely, sorry I missed it on the first pass.

mcmillhj: @rolandjitsu test is now updated.

PR #684, opened 2018-10-05
Added missing semicolon

tvolk131:
The missing semicolon is throwing off tslint in my project build for some reason.

### DIFF
--- a/typings/tests/basic.tsx
+++ b/typings/tests/basic.tsx
@@ -82,4 +82,4 @@ class Basic2 extends React.Component {
 }
 
 // verify that all props are optional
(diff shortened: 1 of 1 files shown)


okonet:
Instead we should align TS config with this project's code standard (no semicolons)

rolandjitsu:
You should not be linting dependencies' code. Could be because your tslint config is not ignoring node_modules. Try adding:
```json
"linterOptions":  {
  "exclude": [
    "node_modules/**"
   ]
}
```

There is also https://github.com/palantir/tslint/issues/3623 which reports somewhat similar behaviour.

In any case, I'm closing this as it does not address any specific issue.

Issue #691, opened 2018-10-11
[Feature Request] Add `open()` to child function parameters


MrLeebo:
**Do you want to request a *feature* or report a *bug*?**

- [ ] I found a bug
- [x] I want to propose a feature

**What is the current behavior?**
To programmatically open a Dropzone, you must `ref` the component and call `open()` on it, even when you're using the `children` prop as a function.

```jsx
class App extends React.Component {
  dropzone = React.createRef()

  render() {
    return (
      <Dropzone ref={this.dropzone}>
        {dropProps => (
          <button onClick={() => this.dropzone.current.open()}>Upload</button>

          ...
        )}
      </Dropzone>
    )
  }
}
```

**What is the expected behavior?**
It seems silly to need to `ref` a component that is already providing me with render props. `open()` should be included in the child function parameters.

```jsx
const App = () => (
  <Dropzone>
    {dropProps => (
      <button onClick={() => dropProps.open()}>Upload</button>

      ...
    )}
  </Dropzone>
)
```

**If this is a feature request, what is motivation or use case for changing the behavior?**
Adding `open()` to the render function parameters makes the component more versatile because you won't have to keep a `ref` or convert a functional component into a class component.

**Please mention other relevant information such as the browser version, Operating System and react-dropzone version.**

```
react-dropzone v6.0.4
```

PR #692, opened 2018-10-12
feat: add open() to child func props (#691)

MrLeebo:

- [ ] bugfix
- [x] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [ ] Not relevant

**Summary**

If you are rendering a Dropzone using a child function, you can open the file dialog programmatically without having to also add a ref.

https://github.com/react-dropzone/react-dropzone/issues/691

No

**Other information**
I updated the TypeScript definition, though I've never used TypeScript before so I just guessed on the syntax. Hopefully someone with more experience can take a look at it and see if that is correct?

### DIFF
--- a/examples/FileDialog/Readme.md
+++ b/examples/FileDialog/Readme.md
@@ -1,15 +1,33 @@
-You can programmatically invoke default OS file prompt. In order to do that you'll have to set the ref on your `Dropzone` instance and call the instance `open` method.
-
 **Please note**, that for security reasons most browsers require popups and dialogues to originate from a direct user interaction (i.e. click). If you are calling `dropzoneRef.open()` asynchronously, there’s a good chance it’s going to be blocked by the browser. So if you are calling `dropzoneRef.open()` asynchronously, be sure there is no more than *1000ms* delay between user interaction and `dropzoneRef.open()` call. Due to the lack of official docs on this (at least we haven’t found any. If you know one, feel free to open PR), there is no guarantee that **allowed delay duration** will not be changed in later browser versions. Since implementations may differ between different browsers, avoid calling open asynchronously if possible.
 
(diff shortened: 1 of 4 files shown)


examples/FileDialog/Readme.md
```diff
@@ -1,15 +1,33 @@
...
+    </React.Fragment>
+  )}
+</Dropzone>
+```
+
+##### Open programmatically by Ref
+
 ```
-let dropzoneRef;
+const dropzoneRef = React.createRef();
```
rolandjitsu: We still need to keep the option to invoke `.open()` using a ref as some users may not use the render fn.

So keep the old example and add another to illustrate what you have added.

src/index.js
```diff
@@ -313,7 +314,8 @@ class Dropzone extends React.Component {
         ...this.state,
         isDragActive,
         isDragAccept,
-        isDragReject
+        isDragReject,
+        open: this.open
```
rolandjitsu: Bind the open fn to the context of the dropzone.

examples/FileDialog/Readme.md
MrLeebo: This is the ref example, I updated it to use the modern `React.createRef()` API instead of the callback API (to encourage best practices) but the callback API still works. Are you suggesting they be split up into two different files?

src/index.js
MrLeebo: It is being bound in the constructor.

rolandjitsu: Sorry, I missed that 👍

examples/FileDialog/Readme.md
rolandjitsu: No, I just missed that, it's been a long day 😄

PR #694, opened 2018-10-13
fix: improve onClick and disableClick props (fixes #530)

### LINKED ISSUE #530: Don't open file dialog when default prevented earlier.
- I found a bug
- ~~I want to propose a feature~~

When clicking a button inside a dropzone that "prevented the default behavior", the dropzone still handles the click.

1. Create a `ReactDropzone`, with a button inside. For example, to delete a dropped file (`DeleteFileButton`)
2. Make sure the `onClick` handler of `DeleteFileButton` calls `event.preventDefault()`.
2. Click this `DeleteFileButton`
3. The Open File Dialog is shown.

When calling `event.preventDefault()` inside the `onClick` handler of the `DeleteFileButton`, the `onClick` handler of `ReactDropzone` shouldn't do anything. This can be achieved by:

```js
onClick(evt) {
  if (evt.isDefaultPrevented()) {
    return;
  }

  // rest of onClick handler.
}
```

ibc:
- Make onClick work when disableClick is true
- Check if click event was prevented during onClick

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [x] Not relevant

- [x] Not relevant

**Summary**

This fixes https://github.com/react-dropzone/react-dropzone/issues/530 by making `onClick` prop run even if `disableClick` is set. It also verifies that the click event was not default-prevented during the onClick listener and, if so, it does not open the file selector dialog.

No. It conforms to the existing documentation API.


src/index.js
```diff
@@ -243,12 +243,16 @@ class Dropzone extends React.Component {
...
-        onClick.call(this, evt)
-      }
+    // if onClick prop is given, run it first
+    if (onClick) {
+      onClick.call(this, evt)
+    }
+
+    // if disableClick is not set and the event hasn't been default prefented within
+    // the onClick listener, open the file dialog
+    if (!disableClick && !evt.isDefaultPrevented()) {
```
rolandjitsu: Could you add a test for this change?

ibc: Hi @rolandjitsu, I'm checking `typings/tests` to learn how to add a test, but I do know where the "actions" or "asserts" are written. May I have a short explanation please?

rolandjitsu: @ibc actually the tests are right alongside the source files. So you should be looking at `src/index.spec.js`. I think there are plenty of tests there to illustrate how everything works.

And from the root of the project, after you installed the npm packages, just run `yarn test` or `jest --watch`.

ibc: ok, thanks, will try.

ibc: Here you have the requested tests :)

rolandjitsu: Let's also make sure that coverage doesn't drop. If you run `yarn test`/`npm test`, you should see which lines are not covered (ignore two lines: one in utils#L54 and `setTimeout(this.open.bind(this), 0)` in the main file).

ibc:
Thanks for merging. Could I know when a new NPM version is published?

PR #708, opened 2018-10-25
refactor: remove file preview and close #705

### LINKED ISSUE #705: refactor: remove {preview} from File

- [ ] I found a bug

Currently, we create a `{preview}` property on the File object (unless the user disables that). This can lead to memory leaks if users do not read through the docs carefully.

1.
1.
1.

I think it would be a good idea to remove this property and let the user handle it (it's a matter of just doing `URL.createObjectURL(file)`) for a few reasons:
- users that do not know that `{disablePreview}` has to be added to the dropzone in order to disable preview generation will not end up with mem. leaks
- avoid confusion when preview might mean to a user that we show an actual preview of the file (image, pdf, etc.)
- users have control over creating/revoking the data uris, whereas we do not control the latter

This would be a breaking change for users that use this feature.

@okonet what is your opinion on this?

rolandjitsu:

- [ ] bugfix
- [ ] feature
- [x] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [ ] Not relevant

**Summary**

Remove the preview generation for files.

The dropzone should not be responsible for enhancing the File object with whatever the user needs. The user is free to manipulate the File objects as they wish after the `onDrop()` callback emits them.

Yes. The `{preview}` property on the File object will no longer be generated. And the `{disablePreview}` property on the Dropzone is also removed.

Creating previews can easily be achieved in the `onDrop()` callback:
```js
class MyZone extends React.Component {
  onDrop = files => {
    this.setState({
      files: files.map(file => ({
        ...file,
        preview: URL.createObjectURL(file)
      }))
    });
  }

  cleanup = () => {
    // Revoke data uris when done using the previews
    const {files} = this.state;

    for (const file of files) {
      URL.revokeObjectURL(file.preview);
    }
  }

  render() {
    return (
      <Dropzone onDrop={this.onDrop} />
    );
  }
}
```

**Other information**

@okonet I haven't added any plugins for this as that would have further implications.

We would need to export other things besides the default dropzone, which means we need to change the rollup build so that it doesn't log warnings in two possible ways:

- no default exports, which means that we export the Dropzone as a named export so that users would `import {Dropzone} from 'dropzone'`
- export other things alongside the default Dropzone export and disable the warning in Rollup, but that's a bad practice as mentioned at [rollupjs.org#exporting](https://rollupjs.org/guide/en#exporting)

Furthermore, we still need to ask the users to remember to revoke the data uris. So I think it's better to let the users handle this on theirs side if they need it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,9 @@
 # react-dropzone
 
 [![npm](https://img.shields.io/npm/v/react-dropzone.svg)](https://www.npmjs.com/package/react-dropzone)
-[![Build Status](https://travis-ci.org/react-dropzone/react-dropzone.svg?branch=master)](https://travis-ci.org/react-dropzone/react-dropzone) 
-[![codecov](https://codecov.io/gh/react-dropzone/react-dropzone/branch/master/graph/badge.svg)](https://codecov.io/gh/react-dropzone/react-dropzone) 
-[![OpenCollective](https://opencollective.com/react-dropzone/backers/badge.svg)](#backers) 
+[![Build Status](https://travis-ci.org/react-dropzone/react-dropzone.svg?branch=master)](https://travis-ci.org/react-dropzone/react-dropzone)
+[![codecov](https://codecov.io/gh/react-dropzone/react-dropzone/branch/master/graph/badge.svg)](https://codecov.io/gh/react-dropzone/react-dropzone)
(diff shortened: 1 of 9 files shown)


okonet:
Agreed, but documentation should reflect that. Also please add an example with previews since I believe that’s a very common use case.

rolandjitsu:
@okonet done, I've added an example of how to generate previews.

PR #748, opened 2019-01-25
Make it compatible with React-like libs

### LINKED ISSUE #736: Make it compatible with React-like libs
Latest versions of react-dropzone introduce the use of event.isDefaultPrevented() which is a jquery-specific[1] function and should be replaced with event.defaultPrevented[2].

[1] https://api.jquery.com/event.isDefaultPrevented/
[2] https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented

jschill:
Fixes #736

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -41,6 +41,30 @@ export function allFilesAccepted(files, accept) {
   return files.every(file => fileAccepted(file, accept))
 }
 
+// React's synthetic events has evt.isPropagationStopped,
+// but to remain compatibility with other libs (Preact) fall back
+// to check evt.cancelBubble
+export function isPropagationStopped(evt) {
+  if (typeof evt.isPropagationStopped === 'function') {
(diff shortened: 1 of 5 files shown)


rolandjitsu:
@jschill there's one test failing.

src/utils/index.js
```diff
@@ -41,6 +41,30 @@ export function allFilesAccepted(files, accept) {
   return files.every(file => fileAccepted(file, accept))
 }
 
+// React's synthetic events has evt.isPropagationStopped,
+// but to remain compatibility with other libs (Preact) fall back
+// to check evt.cancelBubble
+export function isPropagationStopped(evt) {
```
rolandjitsu: Please add tests for these 2 functions as well.

jschill:
@rolandjitsu You're right! Sorry about the delay fixing this.

PR #754, opened 2019-02-12
feat: add {disableKeyDown} prop to disable open file dialog on keydown

rolandjitsu:

- [ ] bugfix
- [x] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Not relevant

**Summary**

Allow users to disable opening the file dialog on keydown (space/enter) when the dropzone is in focus using a `{disableKeyDown}` prop on the `<Dropzone>`.

No.

**Other information**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -275,7 +275,8 @@ class Dropzone extends React.Component {
   }
 
   onKeyDown = evt => {
-    const { onKeyDown } = this.props
+    const { onKeyDown, disableKeyDown } = this.props
+
     if (!this.node.isEqualNode(evt.target)) {
       return
     }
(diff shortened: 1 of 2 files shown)


okonet:
I’m wondering if the same can be auchived by `preventDefault()` on the event

rolandjitsu:
@okonet that's is true, but it could be the same for `onClick`, `onFocus` and `onBlur`. So do we also remove the `disableClick` prop and close this PR?

okonet:
I would propose to deprecate the `disableClick` if the same behavior can be aichived with an idiomatic JS and React. After the deprecation I’d add an example on how to aichive this (with any event). We can also simply remove the prop and release a a major breaking version but if there is a way to do so gradually without introducing lots of maintenance I’d go with it. Your call!

rolandjitsu:
@okonet no, I agree. It will require less maintenance to deprecate the prop and just have an example on how to achieve the same result using idiomatic js/react.

Issue #755, opened 2019-02-12
Dragging 


jonahgreenthal:
**Do you want to request a *feature* or report a *bug*?**
- [x] I found a bug
- [ ] I want to propose a feature

**What is the current behavior?**
Dragging an image from another browser window to a Dropzone doesn't work in Chrome.

**If the current behavior is a bug, please provide the steps to reproduce.**

1. Open https://react-dropzone.js.org and scroll to the "Previews" section
1. In another window of the same browser, open an image, e.g. https://www.google.com/images/branding/googlelogo/2x/googlelogo_color_272x92dp.png
1. Drag and drop the image into the Dropzone

**What is the expected behavior?**
Dropzone should process the image and show a preview.

This works in Safari 12.0.3 (14606.4.5), and possibly in other browsers. However, in Chrome 72.0.3626.96 (both Windows and Mac), nothing happens. I believe `isDragActive` never becomes true.

**Please mention other relevant information such as the browser version, Operating System and react-dropzone version.**

rolandjitsu:
@jonahgreenthal that is expected behaviour.

You are not dragging a file, per say, you are dragging an html element (if you look at the data transfer item type, it's `text/html`), and we intentionally ignore those types because this lib only handles files from your system.

I think [react-dnd](https://github.com/react-dnd/react-dnd) might have what you need.

jonahgreenthal:
So the fact that it works in Safari is just a complete coincidence/accident?

rolandjitsu:
That's probably a a bug or Safari reports that as a file type. I will investigate this tomorrow.

rolandjitsu:
That seems to be the case, Safari reports the drag item type as file. I'm working on figuring out what to do about this, but in principle, we'd like to only handle files with this lib and ignore other types (links, html elements, etc.).

jonahgreenthal:
That's unfortunate; it was super useful to me when it worked. I'll look at react-dnd…but if you're willing to make this package support that functionality across browsers, I'd sure be grateful. Thanks!

rolandjitsu:
@jonahgreenthal unfortunately, Safari seems to be the only browser that can get the file data when an image is dropped from a different browser window.

[...]

PR #757, opened 2019-02-13
refactor: deprecate {disableClick}

rolandjitsu:

- [ ] bugfix
- [ ] feature
- [x] refactoring / style
- [ ] build / chore
- [ ] documentation

- [x] Not relevant

- [ ] Not relevant

**Summary**

We're deprecating the `{disableClick}` property as the same result can be achieved using idiomatic js/react.

No.

**Other information**

### DIFF
--- a/examples/Events/README.md
+++ b/examples/Events/README.md
@@ -1,0 +1,29 @@
+If you'd like to prevent the default behaviour for `onClick`, `onFocus`, `onBlur` and `onKeyDown`, use the `preventDefault()` fn on the event:
+
+```jsx harmony
+class Events extends React.Component {
+  onClick(evt) {
+    evt.preventDefault();
+  }
+
+  render() {
+    return (
+      <Dropzone
+        onClick={this.onClick.bind(this)}
+      >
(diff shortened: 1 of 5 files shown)


rolandjitsu:
@okonet it seems that `refactor` commits do not get released. What should we do about this?

okonet:
Yeah it should have been a feature one. You could force a release with an empty commit I think. Not sure tbh

rolandjitsu:
@okonet I guess we'll release this with the next bug fix. I'm working on figuring out what to do about #755 as it seems Safari is reporting the drag item type as file, so there will be a PR soon.

goodoldneon:
The default prop for `disableClick` is `false`, which means the deprecation warning appears unless `disableClick` is explicitly set it to `null`. I'd think a deprecation warning shouldn't appear if the deprecated prop isn't specified. Is this behavior expected?

rolandjitsu:
@goodoldneon You're right, it should not appear unless the user uses it. But this prop will be removed and it should not be the case anymore.

PR #758, opened 2019-02-15
feat: use file-selector as default file aggregator

rolandjitsu:

- [ ] bugfix
- [x] feature
- [ ] refactoring / style
- [ ] build / chore
- [x] documentation

- [ ] Not relevant

- [ ] Not relevant

**Summary**

Use the [file-selector](https://github.com/react-dropzone/file-selector) plugin by default to aggregate files. This adds folder support by default and removes the necessity of maintaining another fn to aggregate files.

Also:
- remove {onDragStart} prop (only called if the dropzone itself is dragged)
- remove dead code (styles.js)

No.

**Other information**


src/index.js
```diff
@@ -10,7 +11,6 @@ import {
   allFilesAccepted,
   fileMatchSize,
   onDocumentDragOver,
-  getDataTransferItems as defaultGetDataTransferItem,
```
okonet: This removes an ability to plug different file parsers. Why?

src/index.js
```diff
@@ -544,11 +533,6 @@ Dropzone.propTypes = {
    */
   onDropRejected: PropTypes.func,
 
-  /**
-   * onDragStart callback
-   */
-  onDragStart: PropTypes.func,
-
```
okonet: This is a breaking change

okonet:
I’m not sure about this one. It says it s a feature release although it removes 2 features. It also says it’s a non-breaking change there both removal change the API and this must be released as a major version. 

Moreover, I don’t understand the motivation of doing so. Can you please explain it? 

Also it feels like these changes should have been split into 2 separate PRs. Right now I’m now sure what change is affect but what prop removal.

src/index.js
rolandjitsu: It is not, from a functionality pov, because nobody can use this prop. It never gets called unless the dropzone itself is being dragged.

But syntacticly speaking, you're right, it is a breaking change.

src/index.js
rolandjitsu: It does not. Users can still use their own thing. The only difference is that now we use the file-selector fn by default.

rolandjitsu:
> I’m not sure about this one. It says it s a feature release although it removes 2 features. It also says it’s a non-breaking change there both removal change the API and this must be released as a major version.

It is a feature because we add folder support by default. Users do not have to import another library to be able to drag 'n' drop folders anymore.

But true, even though removing `onDragStart` should have no impact on any other apps, ppl might have added this prop anyway.

I will add a breaking change comment in the commit.

> Moreover, I don’t understand the motivation of doing so. Can you please explain it?

First of all, `onDragStart` is never being called, because this event is only triggered on elements that are dragged. And that is not the case for the dropzone.

Secondly, I added folder support by using the file-selector lib so that users do not have to do it themselves.

Lastly, users can still provide their own file aggregator, that feature hasn't changed. Both tests and examples are still there.

> Also it feels like these changes should have been split into 2 separate PRs. Right now I’m now sure what change is affect but what prop removal.

Maybe. Or I could have 2 separate commits instead of 2 PRs.

src/index.js
okonet: It is a breaking change if the api has changed.

okonet:
> Or I could have 2 separate commits instead of 2 PRs.

I also would do two commits if I’d be working alone. But putting two commits into one PR makes it harder to review and harder to understand the motivations.

rolandjitsu:
Ok, I'll make 2 PRs 😄

rolandjitsu:
Closing this, will open separate PRs.

PR #760, opened 2019-02-15
fix: remove {onDragStart} prop

rolandjitsu:

- [x] bugfix
- [ ] feature
- [x] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Not relevant

**Summary**

{onDragStart} was removed as it is never invoked during the drag 'n' drop of files. This event is only triggered for elements that are being dragged.

Yes. The `{onDragStart}` property is removed.
It will only affect users that added this cb by mistake as in practice it is never invoked.

**Other information**
Using `refactor` type in commit does not seem to make a release, so I used the `fix` type.

### DIFF
--- a/examples/Nesting/Readme.md
+++ b/examples/Nesting/Readme.md
@@ -63,7 +63,6 @@ class NestedDropzone extends React.Component {
       <section>
         <div className="dropzone">
           <Dropzone
-            onDragStart={this.createDragHandler('dragstart', 'parent')}
             onDragEnter={this.createDragHandler('dragenter', 'parent')}
             onDragOver={this.createDragHandler('dragover', 'parent')}
             onDragLeave={this.createDragHandler('dragleave', 'parent')}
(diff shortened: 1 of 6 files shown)


okonet:
I’m not sure about this PR. Are you sure this code isn’t executed? I think it’s here to allow more control in the userland but I’m not sure tbh.

rolandjitsu:
@okonet From [dragstart](https://developer.mozilla.org/en-US/docs/Web/Events/dragstart), this event is either triggered for an element that is being dragged, or if a child of the element that has the event listener is dragged. Either way, for a dropzone, this will never be the case.

Furthermore, since the dropzone is now headless and we do not have any logic that affects the functionality of this lib in the cb, users can simply attach the `onDragStart` to the element they render as root:

```jsx
<Dropzone>
    {({getRootProps}) => <div {...getRootProps({onDragStart: evt => console.log(evt)})}/>}
</Dropzone>
```

okonet:
Yes I see your point here but by looking at the code I see we are doing additional checks if we are dragging files etc which makes me suspicious. If you’re sure it is not being used please feel free to proceed. Are all related tests removed as well?

rolandjitsu:
@okonet maybe we need some more users to test this and confirm. So I won't merge this PR for now.

@rxmarbles can you also check that `{onDragStart}` cb is never called when drag 'n' drop of files occurs?

rxmarbles:
sure I should be able to get to that this weekend. if that's ok? just got tied up this week with work

PR #761, opened 2019-02-15
feat: add support for folder drag 'n' drop

rolandjitsu:

- [ ] bugfix
- [x] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [ ] Not relevant

**Summary**

Add support for folder drag 'n' drop by default, no need to install a different library. It should have no effect on users that use their own impl. for file aggregation.

No.

**Other information**


src/index.spec.js
```diff
@@ -1202,6 +1208,7 @@ describe('Dropzone', () => {
         </Dropzone>
       )
       await dropzone.simulate('drop', createDtWithFiles(files))
+      await flushPromises(dropzone)
```
okonet: Question: why is this needed and can’t we put it into `afterEach` for example?

rolandjitsu: I think it was introduced by you or someone else. Without it, tests will fail. It's probably because of the file aggregation which happens async.

okonet: Yes it’s needed because of async get items. But the question still stands: can we remove the repetition and put it into `afterEach`?

rolandjitsu: Will try it and update if it does.

rolandjitsu: Nope, it does not work if I move `flushPromises()` to `afterEach`.

rolandjitsu: @okonet so, what do you think, should we merge this?

okonet:
LGTM please merge and release as a breaking change.

rolandjitsu:
Hmm, it did not release as breaking change. Maybe it's because I did not add the colons after `BREAKING CHANGE`?

okonet:
Oh I’m so sick of how strict and fragile the semantic-release is :(

rolandjitsu:
Should be fixed with https://github.com/react-dropzone/react-dropzone/pull/767.

PR #771, opened 2019-03-01
Allow a composite component to be a direct child

sstur:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Not relevant

**Summary**

The problem is that you can't add a composite component as a direct child (render prop) of the Dropzone component. The existing code assumes a primitive component (div, p, ...etc) and thus treats it's instance (using ref) as a dom element. This precludes the use of something like [React Native Web](https://github.com/necolas/react-native-web) or even a simple hand-rolled component.

There is no breaking change introduced by this PR

**Other information**

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,8 @@
   ],
   "license": "MIT",
   "peerDependencies": {
-    "react": ">=0.14.0"
+    "react": ">=0.14.0",
+    "react-dom": ">=0.14.0"
   },
   "dependencies": {
     "attr-accept": "^1.1.3",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -5,6 +5,7 @@ const uglify = require('rollup-plugin-uglify')
 
 const umdGlobals = {
   react: 'React',
+  'react-dom': 'ReactDOM',
   'prop-types': 'PropTypes'
 }
 
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,7 @@
 /* eslint prefer-template: 0 */
 
 import React from 'react'
+import { findDOMNode } from 'react-dom'
 import { fromEvent } from 'file-selector'
 import PropTypes from 'prop-types'
 import {
@@ -360,12 +361,14 @@ class Dropzone extends React.Component {
     }
   }
 
-  setNodeRef = node => {
-    this.node = node
+  setNodeRef = ref => {
+    // eslint-disable-next-line react/no-find-dom-node
+    this.node = ref ? findDOMNode(ref) : undefined
   }
 
-  setInputRef = input => {
-    this.input = input
+  setInputRef = ref => {
+    // eslint-disable-next-line react/no-find-dom-node
+    this.input = ref ? findDOMNode(ref) : undefined
   }
 
   /**


sstur:
Note: there doesn't seem to be any relevant tests to update.

Note2: there seems to be failing tests on master, unrelated to this PR.

screen shot 2019-02-28 at 22 47 27

rolandjitsu:
@sstur lets hold off with this until I will make a PR for upgrading to hooks, we might just not need this at all.

rolandjitsu:
Furthermore, [findDOMNode](https://reactjs.org/docs/react-dom.html#finddomnode) is deprecated in strict mode.

sstur:
"upgrading to hooks" 🎉🎉🎉Awesome!

I do know that findDOMNode is deprecated, but there isn't really a more elegant solution to this kind of imperative DOM stuff. Considering findDOMNode doesn't have any performance implications it seemed like a good tradeoff to get a working PR for this fix now and then deal with finding a viable replacement for findDOMNode when such a thing makes sense. (If you know of a viable alternative for findDOMNode for this use case, please do let me know)

For the moment, for me to get Dropzone to work with react-native-web projects I basically need to fork this repo and make the above changes because I don't think this can be worked around in userland (correct me if I'm wrong).

sstur:
One random thought in regards to a more elegant solution than findDOMNode.

What if we use createRef and pass that ref to the render function for the consumer of this library to decide how to get it to the dom node.

Example:
```js
<Dropzone onDrop={this.onDrop}>
  {({getRootProps, rootRef, isDragActive}) => {
    return (
      <MyAwesomeComponent
        {...getRootProps()}
        domRef={rootRef}
      >
        {
          isDragActive ?
            <p>Drop files here...</p> :
        }
      </MyAwesomeComponent>
    )
  }}
</Dropzone>
```

In the above case, `MyAwesomeComponent` would be responsible for passing the `rootRef` to the appropriate underlying dom element.

This is a bigger change, but it would avoid findDOMNode. Thoughts?

rolandjitsu:
> I do know that findDOMNode is deprecated, but there isn't really a more elegant solution to this kind of imperative DOM stuff. Considering findDOMNode doesn't have any performance implications it seemed like a good tradeoff to get a working PR for this fix now and then deal with finding a viable replacement for findDOMNode when such a thing makes sense. (If you know of a viable alternative for findDOMNode for this use case, please do let me know)

I see. I don't fully understand what the issue is so I cannot really have an opinion on this. Though, you can use `refKey` (`getRootProps({refKey: 'domRef'})`) to customize how the ref is set, but I suspect that this won't help.

I will try react-native-web to see what the problem is.

> For the moment, for me to get Dropzone to work with react-native-web projects I basically need to fork this repo and make the above changes because I don't think this can be worked around in userland (correct me if I'm wrong).

Yes, that's not acceptable. I am confident we can fix this on our side.

> What if we use createRef and pass that ref to the render function for the consumer of this library to decide how to get it to the dom node.

Well, the PR with the hooks uses `createRef()`, so it might be an option if there isn't a more elegant way.

> The obvious big downside to this is that it is a breaking change to the public API of react-dropzone and for something that can easily be accomplished with findDOMNode.

Perhaps there's a way to be backwards compatible.

sstur:
> "you can use `refKey (getRootProps({refKey: 'domRef'}))` to customize how the ref is set"

This might actually be helpful in producing a workaround, I'm exploring that possibility now.

> "I don't fully understand what the issue is so I cannot really have an opinion on this."

I'll try to provide you a minimal reproducible case on CodeSandbox. This will help me better define the issue also.

I'll post back here with what I come up with.

sstur:
Here's a minimal reproducible case on code sandbox:
https://codesandbox.io/s/13q0k860p3

Note that if you open your dev console and try to drag a file over the dropzone, you'll see this exception:
screen shot 2019-03-04 at 13 02 02

This test case shows the problem without using React-Native-Web (to better keep the test case minimal and reduce potential complexity).

A reasonable initial reaction might be that "MyView" component should allow [ref passthrough](https://reactjs.org/docs/forwarding-refs.html), but keep in mind that in real-life it's likely that MyView component is provided by a library (such as RNW) and not written by us, so we have little control over what it does with refs.

rolandjitsu:
Using a functional component for `MyView` would actually solve the problem, but you're right, this composite comp. might be from some other lib.

I just added a test case and I'm trying to find a solution for it, hopefully by not using `findDOMNode` as for ppl using strict mode it might be a problem.

rolandjitsu:
The thing is, we cannot use `findDOMNode()` as it will definitely be problematic for users using strict mode, and it won't really work when users use `<Fragment>` and return more than 1 node.

So the proper solution would be to expose a ref to the DOM node from within the composite component and tell dropzone what key to look for:
```jsx
class MyView extends Component {
  render() {
    const { children, innerRef, ...otherProps } = this.props;
    return (
      <div ref={innerRef}  {...otherProps}>
        <span>{children}</span>
      </div>
    );
  }
}

<Dropzone>
  {({ getRootProps }) => (
    <MyView {...getRootProps({ refKey: "innerRef" })}>
      Or Drop Here
    </MyView>
  )}
</Dropzone>
```

See [l50rj41k7m](https://codesandbox.io/s/l50rj41k7m) for a working example.

Of course, using a functional component would avoid all these issues (and there's no reason not to use one as react provides hooks for managing state and life cycle):
```jsx
function MyView(props) {
  const { children, ...otherProps } = props;
  return (
    <div  {...otherProps}>
      <span>{children}</span>
    </div>
  );
}
```

And lib maintainers should handle the scenario where ppl want to get a hold of the DOM ref for the component they build (e.g. [material-ui](https://github.com/mui-org/material-ui/blob/129daff17533736f112172d9195625bed43c8235/packages/material-ui/src/utils/withForwardedRef.js#L25) uses `innerRef` to set DOM refs and provide convenience components to use `ref` directly).

Bottom line is, I don't really see a way we can fix this from our side without introducing an unnecessary breaking change.

sstur:
> "we cannot use findDOMNode() as it will definitely be problematic for users using strict mode"

Seems that this is a blocker, and likely not much we can do about it.

> "the proper solution would be to expose a ref to the DOM node from within the composite component and tell dropzone what key to look for"

Yes, this is ideal.

> "lib maintainers should handle the scenario where ppl want to get a hold of the DOM ref"

Agreed, this makes sense. I'll see if I can get [react-native-web](https://github.com/necolas/react-native-web) to add support for such a thing, although I suspect there are other libs/use-cases where we don't have the luxury of a perfect API.

> "Of course, using a functional component would avoid all these issues"

I'm not entirely sure this is true. I don't believe that functional components automatically forward a `ref` (see screenshot A below, taken from [this](https://codesandbox.io/s/6wxrwm59kk)). In fact, I think the same solution you mentioned above ("the proper solution would be to expose a ref to the DOM node") applies to a functional component equally as to a class component (please correct me if I'm wrong).


For now, I'll consider this an unsolvable problem, without the library authors providing the appropriate API for ref forwarding. I'll post back here if I can think of any universally acceptable solution. Thanks again for looking into this!

Screenshot A:
screen shot 2019-03-05 at 11 08 47

PR #772, opened 2019-03-01
feat: migrate to hooks API

rolandjitsu:

- [ ] bugfix
- [x] feature
- [ ] refactoring / style
- [x] build / chore
- [x] documentation

- [ ] Not relevant

- [ ] Not relevant

**Summary**

Now that [hooks](https://reactjs.org/docs/hooks-intro.html) have been officially released for awhile, I've migrated the current codebase to use hooks and made the dropzone itself a hook.

Note that for users that still use class components, we still provide a `<Dropzone>` component, which is just a wrapper around the dropzone hook.

React and react-dom `>= 16.8` are required.
`{getDataTransferItems}` property has been renamed to `{getFilesFromEvent}`.

**Other information**


rolandjitsu:
Great. Try it out and let me know how it feels.

rolandjitsu:
Note to myself, use `useReducer()` for managing the state returned by the hook instead of using `unstable_batchedUpdates()` to batch state updates.

And add a note about upgrading the build deps in the commit message.

okonet:
It looks good to me. Some feedback:

1. Let's not use shortcuts for common words. I noticed usage of `cb` instead of `callback` in the documentation. Also in code, I'd prefer using `event` instead of `evt` to remove any ambiguity
1. Generic examples could you some styles. Right now it's hard to tell if they work.
1. Just by looking at code I doubt there is a lot of value to use hooks for the dropzone but since it's a hyped trend right now I guess people will demand it anyways :)

I think it's good to merge after 1 is resolved. 2 can be done later and 3 is just an opinion ;)

examples/events/README.md
```diff
@@ -0,0 +1,295 @@
...
+  })
+  
+  return rootProps
+}
+
+
+<Parent />
+```
+
+Note that the sort of behavior illustrated above can lead to some confusion. For example, the `onDragLeave` cb of the parent will not be called if the cb for the same event is invoked on the child. This happens because we invoked `stopPropagation()` on the event from the child.
```
okonet: `cb` -> `callback`

rolandjitsu:
@okonet got it. Just did 1. 2 I guess so, I will do it later, unless someone else wants to do it. I agree with 3, but let's give ppl the option to use either way.

PR #779, opened 2019-03-07
fix: check if document exists and fix #775

### LINKED ISSUE #775: document is not defined

- [ x] I found a bug

Version 10.0.0 cause a document undefined error with server side rendering.

1. Enable application with SSR
1. Use react-dropzone
1.

no undefined error

Most likely caused from changing /src/utils/index.js

from
```
export const supportMultiple =
  typeof document !== 'undefined' && document && document.createElement
    ? 'multiple' in document.createElement('input')
    : true
```

to

```
export const supportMultiple = 'multiple' in document.createElement('input')
```

rolandjitsu:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [x] Not relevant

- [x] Not relevant

**Summary**

Check if the `document` object exists to avoid errors in SSR.
@chill5018 I am still getting some errors after fixing this (`n` is not a function). I will need to investigate further.

No.

**Other information**
I am still working on this PR and will try to find why the other errors occur when using SSR.

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -1,6 +1,7 @@
 import accepts from 'attr-accept'
 
-export const supportMultiple = 'multiple' in document.createElement('input')
+export const supportMultiple =
+  typeof document !== 'undefined' && 'multiple' in document.createElement('input')
 
 // Firefox versions prior to 53 return a bogus MIME type for every file drag, so dragovers with
 // that MIME type will always be accepted
--- a/src/utils/index.spec.js
+++ b/src/utils/index.spec.js
@@ -165,6 +165,4 @@ describe('supportMultiple', () => {
     const utils = await import('./index')
     expect(utils.supportMultiple).toBe(false)
   })
-
-  // TODO: Test if no doc
 })


src/utils/index.js
```diff
@@ -1,6 +1,7 @@
 import accepts from 'attr-accept'
 
-export const supportMultiple = 'multiple' in document.createElement('input')
```
KaiHotz: Wouldn't  it be enough  to just change 

`document.createElement('input')`

 to

`React.createElement('input')`  ?

sstur: No, a React element won't really tell you what the browser is capable of, it's just a plain object afterall. The result of `React.createElement('input')` will look like this:
```js
{
  $$typeof: Symbol(react.element)
  props: {}
  type: "input"
}
```

KaiHotz: Ok thanks for the reply

rolandjitsu:
@okonet just tested if the fix works and there seems to be no more problems. Let me know if it's ok to merge.

P.S. I think we should eventually add some integration tests to ensure this won't break in the future.

okonet:
Please take a lead on this project. I’m grounded with kids and other projects atm :)

rolandjitsu:
@okonet sure thing. Good luck 👍

Issue #783, opened 2019-03-09
Regression: Unable to click DOM elements inside drop zone with recommended code


dimroc:
**Do you want to request a *feature* or report a *bug*?**

- [x] I found a bug
- [ ] I want to propose a feature

**What is the current behavior?**

The removal of `disableClick` in https://github.com/react-dropzone/react-dropzone/pull/767 has caused a regression. You are now unable to click on buttons inside a dropzone when performing the recommended alternative because propagation is stopped entirely: 

```javascript
{onClick: evt => evt.preventDefault()} // stops all propagtion
```

`disableClick` skipped the dropzone element, but continued propagation.

The following replacement stops all propagation and is functionally different:

```jsx
return (
      <Dropzone
        accept={accept}
        onDrop={this.onDrop.bind(this)}
      >
        {({getRootProps, getInputProps, isDragActive}) => (
            <div {...getRootProps({onClick: evt => evt.preventDefault()})} style={{position: "relative"}}>
              <input {...getInputProps()} />
```

https://github.com/react-dropzone/react-dropzone/pull/767/files#diff-0b1f9c3babb635d33ee2724a6f8d78b9

**If the current behavior is a bug, please provide the steps to reproduce.**

The Fullscreen example that would reproduce this has been removed in commit https://github.com/react-dropzone/react-dropzone/commit/5398f44f8fdcab8429b0c7769161bba72dad2800#diff-0b1f9c3babb635d33ee2724a6f8d78b9. (Why?)

1. Create a button inside the fullscreen dropzone
2. Disable click on dropzone by following the recommendation
3. Unable to click button.

Or just add a button inside https://github.com/rolandjitsu/react-dropzone/blob/c31e06e93a96162e4c23cd1167270ac61a1bddf5/examples/Fullscreen/Readme.md

**What is the expected behavior?**

I am able to click the button inside the dropzone.

**Please mention other relevant information such as the browser version, Operating System and react-dropzone version.**

1. Broken in v9 and v10, as a result of https://github.com/react-dropzone/react-dropzone/pull/767. 
2. This can be seen because the old logic would continue propagating the click without invoking the file dialog:

```javascript
if (!disableClick && !isDefaultPrevented(evt)) { // I can disable click but still continue propagation
      evt.stopPropagation()
```

3. Would love to see the return of a full screen dropzone example.
4. Let's bring back disableClick, it is functionally different to the current approach.

Thanks for all the cleanup and work!

[...]

PR #801, opened 2019-03-23
feat: add {noClick}, {noKeyboard}, {noDrag} and {noDragEventsBubbling}

rolandjitsu:

- [x] bugfix
- [x] feature
- [ ] refactoring / style
- [ ] build / chore
- [x] documentation

- [ ] Not relevant

- [ ] Not relevant

**Summary**

Bring back the `{disableClick}` prop (with a different name) as using idiomatic js causes some other issues in the user land (e.g. `event.stopPropagation()` will stop events from being propagated to children, see #783).

Alongside that, also introduce:
- `{noKeyboard}` to disable keyboard events (SPACE/ENTER to open file dialog, focus/blur state)
- `{noDrag}` to disable drag events
- `{noDragEventsBubbling}` to stop drag events propagation to parents

No.

**Other information**

### DIFF
--- a/examples/events/README.md
+++ b/examples/events/README.md
@@ -1,160 +1,9 @@
-Custom event handlers provided in `getRootProps()` (e.g. `onClick`), will be invoked before the dropzone handlers.
-
-Therefore, if you'd like to prevent the default behaviour for: `onClick` and `onKeyDown` (open the file dialog), `onFocus` and `onBlur` (sets the `isFocused` state) and drag events; use the `stopPropagation()` fn on the event:
-
-```jsx harmony
-import React, {useCallback, useReducer} from 'react';
-import {useDropzone} from 'react-dropzone';
-
-const initialEvtsState = {
-  preventFocus: true,
-  preventClick: true,
-  preventKeyDown: true,
-  preventDrag: true,
-  files: []
-};
-
-function Events(props) {
-  const [state, dispatch] = useReducer(reducer, initialEvtsState);
-  const myRootProps = computeRootProps(state);
-  const createToggleHandler = type => () => dispatch({type});
-
-  const onDrop = useCallback(files => dispatch({
-    type: 'setFiles',
-    payload: files
-  }), []);
-
-  const {getRootProps, getInputProps, isFocused} = useDropzone({onDrop});
-  const files = state.files.map(file => <li key={file.path}>{file.path}</li>);
-
-  const options = ['preventFocus', 'preventClick', 'preventKeyDown', 'preventDrag'].map(key => (
-    <div key={key}>
-      <input
-        id={key}
-        type="checkbox"
-        onChange={createToggleHandler(key)}
-        checked={state[key]}
-      />
-      <label htmlFor={key}>
-        {key}
-      </label>
-    </div>
-  ));
-
-  return (
-    <section>
-      <aside>
-        {options}
-      </aside>
-      <div {...getRootProps(myRootProps)}>
-        <input {...getInputProps()} />
-        <p>{getDesc(state)} (<em>{isFocused ? 'focused' : 'not focused'}</em>)</p>
-      </div>
-      <aside>
-        <h4>Files</h4>
-        <ul>{files}</ul>
-      </aside>
-    </section>
-  );
-}
-
-function computeRootProps(state) {
-  const props = {};
-
-  if (state.preventFocus) {
-    Object.assign(props, {
-      onFocus: event => event.stopPropagation(),
-      onBlur: event => event.stopPropagation()
-    });
-  }
-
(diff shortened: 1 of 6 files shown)


examples/events/README.md
```diff
@@ -292,4 +111,129 @@ function computeDropzoneProps(props, node) {
...
+
+  const onDrop = useCallback(files => dispatch({
+    type: 'setFiles',
+    payload: files
+  }), []);
+
+  const {getRootProps, getInputProps, isFocused} = useDropzone({...state, onDrop});
+  const files = state.files.map(file => <li key={file.path}>{file.path}</li>);
+
+  const options = ['noClick', 'noKeyboard', 'noDrag'].map(key => (
```
dimroc: While I appreciate the sophisticated use of react hooks like `useCallback` and the meta programming on the options, I think the goal of a README.md is readability. Could we emphasize a simple use case and allow readers to extrapolate from that one scenario to all other options? I believe you touched on this with a `noDragEventsBubbling` option.

examples/events/README.md
```diff
@@ -221,8 +46,10 @@ function Parent(props) {
 }
 
 function Child(props) {
-  const dropzoneProps = useMemo(() => computeDropzoneProps(props, 'child'), [
-    props.preventDrag,
+  const dropzoneProps = useMemo(() => ({
+    ...computeDropzoneProps(props, 'child'),
+    noDragEventsBubbling: true
+  }), [
```
dimroc: This is the simplest use of the new options in the README.md. Will help at least 90% of users that read this. Had to grok a lot of this to realize the `noClick` option is analogous to this.

examples/events/README.md
```diff
@@ -292,4 +111,129 @@ function computeDropzoneProps(props, node) {
...
+};
+
+function Events(props) {
+  const [state, dispatch] = useReducer(reducer, initialEvtsState);
+  const createToggleHandler = type => () => dispatch({type});
+
+  const onDrop = useCallback(files => dispatch({
+    type: 'setFiles',
+    payload: files
+  }), []);
```
dimroc: Since you're not listening to changes on any dependencies, isn't this overkill? Could this be rewritten as a simple function with type defined in the ts definition: 

`onDrop?(acceptedFiles: File[], rejectedFiles: File[], event: DropEvent): void`

```
const onDrop = files => dispatch({
    type: 'setFiles',
    payload: files
  });
```

My overarching goal here is simplicity and readability for future programmers. Even if they're adept at react hooks, why parse complexity if it could be simple.

examples/file-dialog/README.md
```diff
@@ -11,19 +11,14 @@ import React from 'react';
...
+  const {getRootProps, getInputProps, open} = useDropzone({
     // Disable click and keydown behavior
-    onClick: event => event.stopPropagation(),
-    onKeyDown: event => {
-      if (event.keyCode === 32 || event.keyCode === 13) {
-        event.stopPropagation();
-      }
-    }
+    noClick: true,
+    noKeyboard: true
```
dimroc: 👍

examples/file-dialog/README.md
```diff
@@ -43,26 +38,24 @@ import React, {createRef} from 'react';
...
+  // Note that the ref is set async,
+  // so it might be null at some point 
+  if (dropzoneRef.current) {
+    dropzoneRef.current.open()
+  }
+};
 
-<Dropzone ref={dropzoneRef}>
+// Disable click and keydown behavior on the <Dropzone>
+<Dropzone ref={dropzoneRef} noClick noKeyboard>
```
dimroc: 👍

examples/events/README.md
rolandjitsu: I can definitely simplify some of the examples and perhaps create some codesandbox examples and just link to them instead.

examples/events/README.md
rolandjitsu: Will try to update the other examples to match this one.

examples/events/README.md
rolandjitsu: Using the `useCallback(..., [])` hook that way creates a memoized version of the fn, which means we avoid re-rendering (because internally `useDropzone()` also caches and has `onDrop` as dependency).

But I guess it's not that important and we could remove it.

rolandjitsu:
> I left some optional feedback regarding simplicity and readability. I got consumed with the `examples/events/README.md` only realize there are more focused examples further down.
> 
> Thanks for the work!

Thanks, I appreciate you took the time to leave some feedback. I will make the necessary changes over this weekend.

rolandjitsu:
I will actually merge this as it is and update the examples to be simpler in a different PR and will also style some of them as it's hard to do drag 'n' drop in some cases (where there's just the text).

PR #835, opened 2019-05-21
fix: change typescript types to FileWithPath instead of File

majames:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [x] Not relevant

- [x] Not relevant

**Summary**

Small change to type definitions of `files` from `File` to `FileWithPath`.

Motivation is that I receive type errors when trying to access the `path` property in the `acceptedFiles` array returned from `onDrop` as they are incorrectly typed as `File`.

No, as `FileWithType` extends `File` (only difference is optional `path` property)

### DIFF
--- a/typings/react-dropzone.d.ts
+++ b/typings/react-dropzone.d.ts
@@ -1,5 +1,6 @@
 import * as React from "react";
 
+export { FileWithPath } from "file-selector";
 export default function Dropzone(props: DropzoneProps & React.RefAttributes<DropzoneRef>): JSX.Element;
 export function useDropzone(options?: DropzoneOptions): DropzoneState;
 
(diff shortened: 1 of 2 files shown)


rolandjitsu:
The problem with doing this is that if ppl use other sort of plugins, there will be no `{path}` property. So actually, the type of the `acceptedFiles`/`rejectedFiles` would be `T extends File` where `T` can by any type that is a superset of `File`.

I think a better solution is to make the callback's accept a generic type:
```ts
onDrop<T extends File>(acceptedFiles: T[], ...)
```

Then in your code you can just typecast:
```ts
<Dropzone onDrop={onDrop}>

function onDrop(acceptedFiles: FileWithPath, ...) {
  ...
}
```

Note that the above is pseudo-code, so I'm not sure if it will work without changes.

majames:
Thanks for the quick reply @rolandjitsu! I totally agree that a generic type is more flexible and have updated the PR with your suggestion (accidentally amended commit)

typings/react-dropzone.d.ts
```diff
@@ -1,5 +1,6 @@
 import * as React from "react";
 
+export { FileWithPath } from "file-selector";
```
rolandjitsu: I'm not so sure about this 🤔 Ppl could import it from the file-selector lib as it's already a dependency of this lib (I think).

rolandjitsu:
I actually prefer amend instead of multiple commits, I would have squashed them anyway.

typings/react-dropzone.d.ts
majames: Technically, consumers _could_ import `FileWithPath` from `file-selector`... however, there'd be no safe guard for us from removing the `file-selector` dependency from future minor/patch versions of `react-dropzone` and breaking said consumers.

This leaves us with two options:

1. Export the type from `react-dropzone` (as done here)
2. Move the `file-selector` dep to a `peerDependency` and `devDependency`

I much prefer the first as it less work for consumers of the library to get setup and peer deps can be confusing 🙂

PR #855, opened 2019-07-24
Enable disabling disabled

chadoh:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [x] Not relevant

- [x] Not relevant

**Summary**

If someone sets `disabled: false`, they want the input to be _not disabled_. That is, they want it to be enabled.

But `false || noClick` always returns `noClick`, so the `noClick` setting wins.

Without this change, there's a bug:

* set `noClick: true`
* set `onClick={open}` on a button somewhere
* since the input is disabled, clicking it does nothing, and the file picker cannot be opened

Code to create this situation:

```js
const { getRootProps, getInputProps, open } = useDropzone({ noClick: true })

return (
  <div {...getRootProps()}>
    <Button onClick={open}>Upload</Button>
    <input {...getInputProps({ disabled: false })} />
  </div>
)
```

This raises another question, though: why do we need that `disabled: false` at all? Given I want to disable `onClick` on that wrapping `div`, why would we ever need or want to set `disabled` on the input? I want to remove an `onClick`, not disable my file input altogether.

No.

**Other information**

Bug [introduced here](https://github.com/react-dropzone/react-dropzone/commit/a159354d6281ea2a2cd92f1b1d4a5510412a1fe2#diff-1fdf421c05c1140f6d71444ea2b27638R719)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -716,7 +716,7 @@ export function useDropzone({
         onClick: composeHandler(composeEventHandlers(onClick, onInputElementClick)),
         autoComplete: 'off',
         tabIndex: -1,
-        disabled: disabled || noClick,
+        disabled: disabled !== undefined ? disabled : noClick,
         [refKey]: inputRef
       }
 


chadoh:
Thanks!

I still have this question, though:

> why do we need that `disabled: false` at all? Given I want to disable `onClick` on that wrapping div, why would we ever need or want to set `disabled` on the input? I want to remove an `onClick`, not disable my file input altogether.

rolandjitsu:
> Thanks!
> 
> I still have this question, though:
> 
> > why do we need that `disabled: false` at all? Given I want to disable `onClick` on that wrapping div, why would we ever need or want to set `disabled` on the input? I want to remove an `onClick`, not disable my file input altogether.

I'm not sure I understand what you're referring to. But if it is what I think, having disabled on the input fixes the issue when you use a `<label>` as root ref which would trigger the click on the input twice.

chadoh:
To me, the `noClick` option means that _the wrapper should not have a click handler_. It does not mean _the input should be disabled and all means of opening the file picker should be disabled with it_.

If want a functional file selector, but I want it to be triggered by my own button instead of a click on the wrapper, the current design makes me do two things:

1. set `noClick: true`
2. set `disabled: false` on the input

I think only the first should be necessary. 

The "labels can cause double open" issue seems unrelated. The fix for it doesn't seem to me like it should affect this design decision.

PR #878, opened 2019-09-14
fix: `isDragReject` now true if a file is not within size range

### LINKED ISSUE #861: 'isDragReject' not called when excceeded maxSize

- [X] I found a bug

1. Listen for `isDragReject` event and set `maxSize` parameter
2. Drag file in front of drop zone
3. Event `isDragReject` not fire when file is larger than `maxSize` parameter

TheLordHighExecutioner:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Not relevant

**Summary**

Previously, `isDragReject` would not trigger if a file was not within the specified size range. For example, it would not return `true` if a file was too large. See: https://github.com/react-dropzone/react-dropzone/issues/861. This is because it was not checking for size.

No

**Other information**

Fixes #861


TheLordHighExecutioner:
Suggestion: `onDropAccepted` and `isDragAccepted` should probably behave similarly. A file should trigger both or neither, but not only one.

They both rely on the expression `fileAccepted(file, accept) && fileMatchSize(file, maxSize, minSize)`. Thus, I suggest extracting this to its own function (I could include that in this PR).

rolandjitsu:
Please rebase your PR. I will check it soon to verify that this change doesn't break some intended behavior.

TheLordHighExecutioner:
> Please rebase your PR. I will check it soon to verify that this change doesn't break some intended behavior.

Just rebased it!

PR #899, opened 2019-10-23
feat: adding single prop to override multiple

rgibson518:

- [ ] bugfix
- [x] feature
- [x] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Yes, I've updated the documentation*
- [ ] Not relevant

*NOTE: I've updated the documentation around `single`, but not `multiple`. I think if these changes are acceptable, there is a path to deprecate `multiple` down the road, however it would eventually be a breaking change so it probably should not happen in this PR. One path forward might be to encourage usage of the `single` prop and provide warnings when the `multiple` prop is provided by the user.  I'm happy to update the docs to encourage the use of single over multiple if desired.

**Summary**

This allows an opt in shortcut prop of `single` to turn off the
default `multiple` prop.

Example:
```
// shortcut boolean single prop
<DropZone single>...</DropZone>
// versus old default true multiple prop
<DropZone multiple={false}>...</DropZone>
```

This works on the assumption that if a user is providing single, as
opposed to the default true value of multiple, they intend for the
component to work as `multiple={false}`.

If provided BOTH props with a true value, it seems reasonable that
since the default is already `multiple={true}`, the presence of
`single` is a signal that the user intends the component to be single.
Otherwise, there would be no way of discerning if `multiple === true`
is a product of the default value or the user explicitly including it.

**Misc**
Also fixed the name of a test that implies it is testing changes to the `multiple` prop, however it is testing changes to the `accept` prop.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   "size-limit": [
     {
       "path": "dist/index.js",
-      "limit": "7 KB"
+      "limit": "7.03 KB"
     },
     {
       "path": "dist/es/index.js",
@@ -48,7 +48,9 @@
   },
   "jest": {
     "clearMocks": true,
-    "setupFilesAfterEnv": ["<rootDir>/testSetup.js"],
+    "setupFilesAfterEnv": [
+      "<rootDir>/testSetup.js"
+    ],
     "coveragePathIgnorePatterns": [
       "/dist/",
       "/node_modules/",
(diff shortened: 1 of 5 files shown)


rolandjitsu:
I don't see the value of adding this prop. You can achieve the same using the `multiple={false}` prop. And the multiple prop is also familiar to the already existing [multiple](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#attr-multiple) input attribute.

Issue #630, opened 2018-07-17
Officially supported browsers ?


Alekseykua:
- [x] I want to propose a feature

**What is the current behavior?**
Does not specify which browsers are fully/partially supported. 

**If this is a feature request, what is motivation or use case for changing the behavior?**
Before integrating users would like to know if their desired browser support is available. Would also help to build support if we know which browsers need it.

Alekseykua:
Bump

slikts:
Ideally this should be done using [browserslist](https://github.com/browserslist/browserslist).

amanadi007:
Any update on this?

rolandjitsu:
@amanadi007 Not really, sorry. I was planning on running e2e tests on SauceLabs which would have helped, but never got the chance to do it.

PR #942, opened 2020-03-11
feat: adds removeFiles, onRemoveFiles

estrattonbailey:

- [ ] bugfix
- [x] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [ ] Not relevant

**Summary**
I came across a situation where I'd like to be able to remove files from the internal state of `react-dropzone`. I saw that others have asked for this, and that [help was needed](https://github.com/react-dropzone/react-dropzone/issues/805#issuecomment-580034406).

This PR introduces two new API methods `onRemoveFiles` and `removeFiles`:
```jsx
<Dropzone onRemoveFiles={files => console.log('removed', files)}>
  {({ removeFiles, acceptedFiles }) => (
    <button onClick={() => removeFiles(acceptedFiles[0])}>Remove one file</button>
  )}
</Dropzone>
```

`onRemoveFiles` will always receive an array of `File`s, and `removeFiles` always accepts an array of `File`s. This makes it easy to remove multiple files at a time, say to "clear" the dropzone if needed.

Shouldn't.

**Other information**
I could use some feedback on the tests, types, and JSDoc. Generally tried to stick to the patterns in place, but lmk if those need work.

Also, thanks for the library :)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -202,7 +202,14 @@ Dropzone.propTypes = {
    * @param {object[]} files
    * @param {(DragEvent|Event)} event
    */
-  onDropRejected: PropTypes.func
+  onDropRejected: PropTypes.func,
+
+  /**
+   * Cb for when a user removes one or more files.
+   *
+   * @param {object[]} files
+   */
+  onRemoveFiles: PropTypes.func
 }
 
 export default Dropzone
@@ -269,6 +276,7 @@ export default Dropzone
  * @property {File[]} draggedFiles Files in active drag
  * @property {File[]} acceptedFiles Accepted files
  * @property {File[]} rejectedFiles Rejected files
+ * @property {Function} removeFiles Accepts an array of File objects to be removed
  */
 
 const initialState = {
@@ -368,6 +376,7 @@ export function useDropzone({
   onDropAccepted,
   onDropRejected,
   onFileDialogCancel,
+  onRemoveFiles,
   preventDropOnDocument = true,
   noClick = false,
   noKeyboard = false,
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -710,6 +719,26 @@ export function useDropzone({
     [inputRef, accept, multiple, onDropCb, disabled]
   )
 
+  const removeFiles = React.useCallback(
+    files => {
+      const acceptedFiles = state.acceptedFiles.filter(acceptedFile => {
+        return !files.find(f => f === acceptedFile);
```
nylon22: `Array.prototype.find` is not supported in IE: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

This project doesn't have an official list of supported browsers. See #630. That being said, the project isn't using `find` anywhere else, so this would be a breaking change for any project that isn't polyfilling `find` and that has IE users. 

I don't maintain this project but I think it would be better to use `Array.prototype.some` instead. :)

estrattonbailey:
@nylon22 thanks for the feedback.

In interest of time, I've actually just wrapped this library to create an API that's more suitable to my needs. [I made a gist](https://gist.github.com/estrattonbailey/b6912c704454478f57f7cb7657d8ebad) that outlines the approach I took.

Because of that, I'm not going to follow up on this PR, and I think I'll close to help reduce clutter.

Angelk90:
@estrattonbailey : I have a similar problem can I ask you for a hand on something?

RajeshCashapona:
Virat - Copy

Issue #911, opened 2019-11-13
[Firefox] isDragActive is wonky on text


Togrias:
**Do you want to request a *feature* or report a *bug*?**

- [x] I found a bug
- [ ] I want to propose a feature

**What is the current behavior?**
isDragActive returns false when dragging over text. This bug is reproducible on Firefox but seems to be working properly on Chrome.

**If the current behavior is a bug, please provide the steps to reproduce.**

The codesandbox [example](https://codesandbox.io/embed/stupefied-kepler-btkpt) displays this behaviour. Simply drag a file over the text in the example and the border will turn white (instead of green) in some places, especially right above the words.

**What is the expected behavior?**
isDragActive should be true when dragging over any children.

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please mention other relevant information such as the browser version, Operating System and react-dropzone version.**

rolandjitsu:
I think this is related to https://github.com/react-dropzone/react-dropzone/issues/888.

Togrias:
It's a separate issue. This bug occurs even on older versions of react-dropzone (See material-ui-dropzone) where #888 worked properly.

But it might be a bug with Firefox... I can't reproduce it on Chrome or Safari.

rolandjitsu:
Ok, I will reopen. I don't have time to investigate though, so you're welcome to do so and make a PR.

byoungb:
I was able to fix this with CSS in Firefox, if you have no background and add `z-index:-1` to the p tag.

ivzinoviev:
Reproduced in FF 70.0.1. Fixed by removing `line-height` style from text.

byoungb:
> Reproduced in FF 70.0.1. Fixed by removing `line-height` style from text.

Not a bad solution but this is not great when text needs to wrap on smaller screen resolutions.

rolandjitsu:
I can confirm that there seems to be an issue in Firefox, unrelated to the issue I referenced initially. Might have something to do with the way we check if `ondragleave` evt happened on a child of the dropzone.

I'll try to spend a few minutes looking into it this week or over the weekend, but if anyone wants to look into it, feel free do to so and make a PR.

remijean:
Same problem with Chrome 79

depiction:
This is still an issue. The workarounds mentioned above didn't work for me.

Issue #526, opened 2017-10-27
In Firefox isDragActive doesn't work correctly when dragging over child elements


kurtfurbush:
**Do you want to request a *feature* or report a *bug*?**

- [x] I found a bug
- [ ] I want to propose a feature

**What is the current behavior?**
In Firefox, dragging a file and hovering over Dropzone children causes isDragActive to change to false. Hovering over the rest of the Dropzone correctly triggers isDragActive = true.

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Using Firefox, go to the react-dropzone Basic Example.
2. Drag a file over the dropzone below the text (child element).
3. Note the green outline.
4. Drag the file over the text (child element).
5. Note the active styling reverts.

**What is the expected behavior?**
Dragging the file over dropzone children should not cause isDragActive to return false. Works as expected in Chrome, IE, Edge.

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please mention other relevant information such as the browser version, Operating System and react-dropzone version.**
Firefox: 56.0.2
OS: Windows 10

okonet:
Thanks for reporting this one. I can confirm there is a bug. Interestingly we have a code that worked in FF as well so this is a regression (maybe even in Firefox). Can you please submit a PR to fix it?

okonet:
Related code: https://github.com/react-dropzone/react-dropzone/blob/master/src/index.js#L91-L94

kurtfurbush:
I've been trying to debug this for a while but keep getting "too much recursion" debug errors in Firebug.

It seems that the onDragEnter event is possibly not happening before onDragLeave (as it does in Chrome) when moving from parent > child and vice-versa, which means this.dropTargets list doesn't correctly contain the "other" element (pushed in onDragEnter), which causes onDragLeave to continue and set isDragActive to false.

bologer:
Issue is relevant, any ideas how to patch it?

bologer:
`onDragLeave()` is not being fired as well when I move file to dropdown area and then move it away.

andygock:
I'm also experiencing this issue, Firefox 64.0, does anyone have any solutions on this?

cupofjoakim:
I'm currently experiencing this as well on Firefox 67.0.4, any known fixes? :(

remijean:
Même problème sous Chrome 79

cupofjoakim:
@okonet Can we reopen this?

okonet:
[...]

PR #949, opened 2020-03-28
fix: #911 isDragActive value when dragging over text on Firefox

avandecreme:

- [X] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [X] Not relevant

**Summary**

This PR fixes issues #911 and #526 
The issue is that when dragging over an element containing a text node, Firefox fires an event when entering the element and then another one when entering the text node.
This can be seen with this fiddle: https://jsfiddle.net/493wmzvL/2/
firefox-drag

With React the behavior is similar however both the text node and its container fire an event with the same contents in `event.target`. This can be seen in this fiddle: https://jsfiddle.net/wqcsg2un/2/
firefox-react-drag
I don't know if it is expected behavior from React or not, I will ask over there.

My suggested fix is to always add elements when `dragEnter` is fired and remove elements one by one when `dragLeave` is fired.

**Other information**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -487,10 +487,7 @@ export function useDropzone({
       event.persist()
       stopPropagation(event)
 
-      // Count the dropzone and any children that are entered.
-      if (dragTargetsRef.current.indexOf(event.target) === -1) {
-        dragTargetsRef.current = [...dragTargetsRef.current, event.target]
-      }
+      dragTargetsRef.current = [...dragTargetsRef.current, event.target]
 
       if (isEvtWithFiles(event)) {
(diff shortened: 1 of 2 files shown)


avandecreme:
Here is the bug I opened on react: https://github.com/facebook/react/issues/18416

rolandjitsu:
@avandecreme could you rebase this PR onto master and test if your fix still works?

avandecreme:
@rolandjitsu Done. It is still working.

PR #1011, opened 2020-09-05
chore: update deps and fix the act() test warnings

rolandjitsu:

- [ ] bugfix
- [ ] feature
- [ ] refactoring / style
- [x] build / chore
- [ ] documentation

- [x] Not relevant

- [ ] Not relevant

**Summary**

Get rid of the dependency security warnings and of the `act(...)` warnings in tests.

Nope.

**Other information**
Note that we need Node 10+ for any dev on the project (due to rollup), so we have to drop that from the test matrix.


rolandjitsu:
@okonet I think we should either get rid of yarn or npm lock files. Doesn't make much sense to have both and yarn complains about it.

rxmarbles (dismissed):

README.md
```diff
@@ -258,7 +253,7 @@ test('invoke onDragEnter when dragenter event occurs', async () => {
       )}
     </Dropzone>
   )
-  const { container } = render(ui)
+  const { container, rerender } = render(ui)
```
donedgardo, line 256: in line 260 I think you have to change the arguments of flushPromises

donedgardo (changes requested):
I've tried this and now I don't get the act warnings, although the test fail as dragOnEnter jest mock function is not being called.

README.md
rolandjitsu: Thanks for reporting this. I've just updated it.

rxmarbles (approved):

PR #1041, opened 2021-01-26
Feature/custom validator

zVolodymyr:

- [ ] bugfix
- [x] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [ ] Not relevant

**Summary**

I my current project I need to reject files with long names. Since validation rules may change project to project I think it's good idea to have ability to specify custom validation.

No breaking changes

**Other information**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -624,7 +624,7 @@ export function useDropzone(options = {}) {
           files.forEach(file => {
             const [accepted, acceptError] = fileAccepted(file, accept)
             const [sizeMatch, sizeError] = fileMatchSize(file, minSize, maxSize)
-            const customErrors = validator?.(file);
+            const customErrors = validator ? validator(file) : null;
 
             if (accepted && sizeMatch && !customErrors) {
               acceptedFiles.push(file)
--- a/src/index.js
+++ b/src/index.js
@@ -60,7 +60,8 @@ const defaultProps = {
   noClick: false,
   noKeyboard: false,
   noDrag: false,
-  noDragEventsBubbling: false
+  noDragEventsBubbling: false,
+  validator: null
 }
 
 Dropzone.defaultProps = defaultProps
@@ -226,7 +227,14 @@ Dropzone.propTypes = {
    * @param {FileRejection[]} fileRejections
    * @param {(DragEvent|Event)} event
    */
-  onDropRejected: PropTypes.func
+  onDropRejected: PropTypes.func,
+
+  /**
+   * Custom validation function 
+   * @param {File} file
+   * @returns {FileError|FileError[]}
+   */
+  validator: PropTypes.func
 }
 
 export default Dropzone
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -615,11 +624,17 @@ export function useDropzone(options = {}) {
           files.forEach(file => {
             const [accepted, acceptError] = fileAccepted(file, accept)
             const [sizeMatch, sizeError] = fileMatchSize(file, minSize, maxSize)
-            if (accepted && sizeMatch) {
+            const customErrors = validator ? validator(file) : null;
```
rolandjitsu, line 627: Might make sense to check if `validator` is a function?

src/index.js
```diff
@@ -615,11 +624,17 @@ export function useDropzone(options = {}) {
...
+
+            if (accepted && sizeMatch && !customErrors) {
               acceptedFiles.push(file)
             } else {
-              const errors = [acceptError, sizeError].filter(e => e)
-              fileRejections.push({ file, errors })
+              let errors = [acceptError, sizeError];
+              
+              if(customErrors)
+                errors = errors.concat(customErrors);
```
rolandjitsu, line 635:
```suggestion
              if (customErrors) {
                errors = errors.concat(customErrors);
              }
```

typings/react-dropzone.d.ts
```diff
@@ -10,7 +10,7 @@ export interface DropzoneProps extends DropzoneOptions {
 
 export interface FileError {
   message: string;
-  code: "file-too-large" | "file-too-small"|"too-many-files"|"file-invalid-type";
+  code: "file-too-large" | "file-too-small"|"too-many-files"|"file-invalid-type"|string;
```
rolandjitsu, line 13:
```suggestion
  code: "file-too-large" | "file-too-small" | "too-many-files" | "file-invalid-type" | string;
```

rolandjitsu (changes requested):
Could you add an example of this in the examples folder (and add it to the styleguidist config as well)?

src/index.js
zVolodymyr: updated

typings/react-dropzone.d.ts
zVolodymyr: updated

src/index.js
zVolodymyr: I don't think it makes sense. There is TypeProps specified for this property and it should warn about incorrect values.  Also ignoring non-function values is more confusing, than getting "TypeError: ___ is not a function'. Adding such check makes sense only to throw error with better message.

rolandjitsu: I'm not saying to ignore it, I'm saying to maybe throw a more friendly message, but I guess it's ok to leave as is. As for TypeProps, do note that if ppl use this as a hook, they would get no warnings.

zVolodymyr:
Added exmaple

rolandjitsu (approved):

BartusZak:
Is there a possibility to use existing validations and add new ones as well?

I just created a custom validator and `maxFiles` and `maxSize` native validations stopped working.

@VolodymyrBaydalka

PR #1083, opened 2021-08-02
Delete vendor directory

rikschennink:
See: https://github.com/react-dropzone/react-dropzone/issues/1076

**What kind of change does this PR introduce?**

- [ ] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [x] documentation

**Did you add tests for your changes?**

- [ ] Yes, my code is well tested
- [x] Not relevant

**If relevant, did you update the documentation?**

- [ ] Yes, I've updated the documentation
- [x] Not relevant

**Summary**

See: https://github.com/react-dropzone/react-dropzone/issues/1076

**Does this PR introduce a breaking change?**

Probably documentation example will stop working.

Can link to here as an alternative: https://codesandbox.io/s/react-dropzone-pintura-40xh4?file=/src/App.js

PR #1086, opened 2021-08-12
Stop shipping the vendor directory

### LINKED ISSUE #1076: Bug: MIT license, but it includes copyrighted commercial code
The addition of Doka's copyrighted files seems incompatible with the MIT license.

I can see in https://github.com/react-dropzone/react-dropzone/issues/1054 that you said, "Note that the files provided are probably not supposed to be distributed or used without a license. We have them there so that we can illustrate how the integration is done."

The auditors at my company aren't happy with that. It means we're either going to have to fork, or stick to an old version, or find an alternative. All for an "example" of something that we don't even use. I'm sure we're not the only company who will shy away from using the module because of this.

Did the Doka author give you an "ok" to include these files? If so, could you link some sort of disclaimer in the files?

Instead of including the actual files, a short README for Doka users would probably be better.

s100:

- [x] bugfix
- [ ] feature
- [ ] refactoring / style
- [ ] build / chore
- [ ] documentation

- [x] Not relevant

- [x] Not relevant

**Summary**

This fixes #1076. This PR is presented as an alternative to #1083.

`react-dropzone` is apparently intended to be compatible with the Doka image editor ([now called "Pintura" apparently](https://pqina.nl/pintura/)), which explains why it's potentially desirable to have `doka` code in the repository for the purposes of evaluating this compatibility. However, `react-dropzone` should not be redistributing `doka`'s own code, which is released under [this specialist licence](https://pqina.nl/pintura/license/), as `react-dropzone` itself is [released under the MIT licence](https://github.com/react-dropzone/react-dropzone/blob/master/LICENSE), and it's highly unlikely that the two are compatible.

No. The `doka` code is only used at development time, `react-dropzone` does not use it at run time.

**Other information**

N/A

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -6,3 +6,4 @@ styleguide/
 coverage/
 yarn.lock
 .babelrc
+vendor


andresin87 (approved):

rolandjitsu (approved):

rolandjitsu:
Damn it, the commit wasn't following the proper spec. I'll fix this today.

PR #1096, opened 2021-09-24
ref: Replace separate error variables with ErrorCode object

akhmadullin:

- [ ] bugfix
- [ ] feature
- [x] refactoring / style
- [ ] build / chore
- [ ] documentation

- [ ] Not relevant

- [x] Not relevant

**Summary**
It seems a little bit strange, that you store errors as separate variables, that you use in the code and at the same time you have ErrorCode object (aka enum), that has all the same errors, but you don't use it at all.
Снимок экрана 2021-09-24 в 11 09 02

According to the type declaration, error is expected to be of ErrorCode enum type, but in code you use separate strings. It seems to be a little bit confusing, don't you think?
Снимок экрана 2021-09-24 в 11 13 41

So, you use ErrorCode object only for typing, but this typing doesn't describe real code realization.

I suggest to store errors only in the ErrorCode object (as it said in the type declaration) and to use ErrorCode object in the code instead of usage of separate error variables.

Changes from this PR could be breaking for projects, that use imports of separate error variables (that were deleted in this PR).

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -1,16 +1,10 @@
 import accepts from 'attr-accept'
 
-// Error codes
-export const FILE_INVALID_TYPE = 'file-invalid-type'
-export const FILE_TOO_LARGE = 'file-too-large'
-export const FILE_TOO_SMALL = 'file-too-small'
-export const TOO_MANY_FILES = 'too-many-files'
-
 export const ErrorCode = {
-  FileInvalidType: FILE_INVALID_TYPE,
-  FileTooLarge: FILE_TOO_LARGE,
-  FileTooSmall: FILE_TOO_SMALL,
-  TooManyFiles: TOO_MANY_FILES,
+  FileInvalidType: 'file-invalid-type',
+  FileTooLarge: 'file-too-large',
+  FileTooSmall: 'file-too-small',
+  TooManyFiles: 'too-many-files',
 }
 
 // File Errors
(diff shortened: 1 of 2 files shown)


jomarquez21:
If `ErrorCode` is not used the change shouldn't be just to remove it, why hardcode the error codes in an object?

akhmadullin:
Why my changes are hardcode?

I just want type description match with code implementation.

I don't know why you added Error code object few releases ago, while you didn't use it at all.

Firstly, you [changed FileError interface](https://github.com/react-dropzone/react-dropzone/commit/58abc82f05f302dd5227ff7b4e3075c4f4e41e30). 

These changes became a reason of https://github.com/react-dropzone/react-dropzone/issues/1091, because you changed typing, but didn't change the implementation (you didn't use ErrorCode enum in code).

Then you fixed this issue by [adding ErrorCode object](https://github.com/react-dropzone/react-dropzone/commit/2a9ddbb416bbd2c568e2389c082084b86e535c78),  that is not used at all in code, it is just exporting from file.

Ok, you added ErrorCode object, https://github.com/react-dropzone/react-dropzone/issues/1091 disappeared, but now typing also differs from implementation, because you have ErrorCode enum in typing, but in code you use simple strings, not ErrorCode object:

Снимок экрана 2021-09-24 в 11 13 41

134603599-0f265d5a-6725-42f9-a449-8317fab0c96b

Implementation differs from type description.

That's why I suggest to [use ErrorCode object instead of simple strings](https://github.com/react-dropzone/react-dropzone/pull/1096/files).

or

If we want to drop ErrorCode object, we should return union in the FileError interface:

```typescript
type ErrorCode = "file-too-large" | "file-too-small" | "too-many-files" | "file-invalid-type";

export interface FileError {
  message: string;
  code: ErrorCode | string;
}
```

I hope, I clearly stated my idea.

What do you think about it?

Do you see a problem of mismatching type description with implementation?

akhmadullin:
In other words, in current implementation it's pointless to import ErrorCode enum from your lib to use in other projects, because  you don't use it yourself in your code.

jomarquez21:
I understand your point and I do not disagree but what I do not agree with is that the definition is in an object.


I show you why I consider this a mistake.
Captura de Pantalla 2021-10-05 a la(s) 08 54 46

akhmadullin:
Hm, I got your fears.

Yes, ``ErrorCode`` object could be changed, but these changes would be done inside another project, not inside your lib.

It seems, that is not your responsibility and not responsibility of ``react-dropzone``, how lib users can modify ``ErrorCode`` object. I can't imagine any case, when developer would decide to change something, that was imported from another lib. Even if developer do it, it would be his problem, not yours.

As a possible solution, I can suggest to:

- use ``Object.freeze()`` method for making ``ErrorCode`` object immutable;

- stay ``ErrorCode`` as [it is right now in my changes](https://github.com/react-dropzone/react-dropzone/pull/1096/files), without any freezing, because inside ``react-dropzone`` it would not be changed;

- drop ``ErrorCode`` object and return previous type description (use union type instead of enum).

What would you prefer?

jomarquez21:
Maybe this is my personal taste but I consider it correct to keep the variables in a constant, I agree with removing `ErrorCode`

And having said that, I have nothing more to add in this pr.

Thanks for your time.

akhmadullin:
@rolandjitsu, Do you have any thoughts about this PR?

akhmadullin:
@rolandjitsu 👀

rolandjitsu:
For flexibility and backwards compatibility, I'm gonna leave the error code definitions as they are.

PR #1142, opened 2022-02-09
fix: fallback to input click to open the file picker

rolandjitsu:
- [x] Bug Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Style
- [ ] Build
- [ ] Chore
- [ ] Documentation
- [ ] CI

- [ ] Not relevant

- [x] Not relevant

**Summary**
It seems like being in a secure context is not enough to use the FS access API (e.g if triggering from a cross-origin iframe, you get a security error). So this PR checks that and falls back to using the input.

No.

**Other information**
Need to add back the window focus cb in this case so that we call the `onFileDialogCancel` from there 🙄


rolandjitsu:
@rxmarbles @okonet I'm thinking of using alpha/beta/next branches so we can test things out without breaking the main track. I'll have to see what settings are needed for semantic release to work with that.

PR #1159, opened 2022-03-16
Use "KeyboardEvent.key" instead of "Keyboard.keyCode"

skayo:
- [ ] Bug Fix
- [ ] Feature
- [X] Refactoring
- [ ] Style
- [ ] Build
- [ ] Chore
- [ ] Documentation
- [ ] CI

- [ ] Not relevant

- [X] Not relevant

**Summary**
`KeyboardEvent.keyCode` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) and should not be used anymore.  
Browsers could drop support anytime and it's not recommended to use it.

I don't think so.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -743,8 +743,8 @@ export function useDropzone(options = {}) {
       }
 
       if (
-        (event.key === 'Enter' || event.key === ' ') ||
-        (event.keyCode === 32 || event.keyCode === 13)
+        event.key === 'Enter' || event.key === ' ' ||
+        event.keyCode === 32 || event.keyCode === 13
       ) {
         event.preventDefault();
         openFileDialog();
--- a/src/index.js
+++ b/src/index.js
@@ -743,8 +743,10 @@ export function useDropzone(options = {}) {
       }
 
       if (
-        event.key === 'Enter' || event.key === ' ' ||
-        event.keyCode === 32 || event.keyCode === 13
+        event.key === " " ||
+        event.key === "Enter" ||
+        event.keyCode === 32 ||
+        event.keyCode === 13
       ) {
         event.preventDefault();
         openFileDialog();
--- a/src/index.js
+++ b/src/index.js
@@ -742,7 +742,11 @@ export function useDropzone(options = {}) {
         return;
       }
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -742,7 +742,11 @@ export function useDropzone(options = {}) {
         return;
       }
 
-      if (event.keyCode === 32 || event.keyCode === 13) {
+      if (
+        (event.key === 'Enter' || event.key === ' ') ||
+        (event.code === 'Enter' || event.code === 'Space') ||
```
rolandjitsu, line 747: From MDN:
```
For example, the code returned is "KeyQ" for the Q key on a QWERTY layout keyboard, but the same code value also represents the ' key on Dvorak keyboards and the A key on AZERTY keyboards. That makes it impossible to use the value of code to determine what the name of the key is to users if they're not using an anticipated keyboard layout.
```

So `evt.code` is also not reliable. But `evt.key` seems like it is. So I don't think we need this one.

skayo: Alright, just added it cause I wasn't sure

src/index.js
```diff
@@ -742,7 +742,10 @@ export function useDropzone(options = {}) {
         return;
       }
 
-      if (event.keyCode === 32 || event.keyCode === 13) {
+      if (
```
rolandjitsu, line 745: We'll obviously need some unit tests. There's already a bunch of them testing the `keyCode`.

skayo: I'll look into it when I have the time!

skayo:
@rolandjitsu I'm a bit unsure about how exactly we should test this properly. I noticed that most people are using `key` and `keyCode` simultaneously in `fireEvent.keyDown`. For example:
```
fireEvent.keyDown(dropzone, {
  key: 'Enter',
  keyCode: 13,
});
```

Should I also do this? Or write separate test for each `keyCode` and `key`?

skayo:
@rolandjitsu added tests now. Could you take a look?

rolandjitsu:
> @rolandjitsu added tests now. Could you take a look?

Thanks. I'll have a look this weekend.

rolandjitsu (approved):

PR #1172, opened 2022-04-11
Update README.md

Uluwatta:
- [x] Bug Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Style
- [ ] Build
- [ ] Chore
- [ ] Documentation
- [ ] CI

- [x] Not relevant

- [x] Not relevant

**Summary**
Explain the **motivation** for making this change. What existing problem does the pull request solve?
Previous example 'Previews' won't work properly on Chrome, images won't load properly , this is due to the blob URLs are getting revoked before they are being consumed
'files' were used as a dependency on use effect, and since useDropzone uses setFiles which updates the files state will cause blob URLs revoke.

No

**Other information**

**Problem :** Previous example won't work properly on Chrome, images won't load properly , this is due to the blob URLs are getting revoked before they are being consumed
'files' were used as a dependency on use effect, and since useDropzone uses setFiles which updates the files state will cause blob URLs revoke.

**Improvements :** 

1. Revoked blob URLs on image load
2. Revoked blob URLs on component unmount

### DIFF
--- a/examples/previews/README.md
+++ b/examples/previews/README.md
@@ -55,15 +55,17 @@ function Previews(props) {
         <img
           src={file.preview}
           style={img}
+          // Revoke data uri after image is loaded
+          onLoad={() => { URL.revokeObjectURL(file.preview) }}
         />
       </div>
     </div>
   ));
 
   useEffect(() => {
-    // Make sure to revoke the data uris to avoid memory leaks
-    files.forEach(file => URL.revokeObjectURL(file.preview));
-  }, [files]);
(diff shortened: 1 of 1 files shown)


Uluwatta:
Hey folks, is there any feedback on this ?

Uluwatta:
Issue : https://github.com/react-dropzone/react-dropzone/issues/1149

rolandjitsu (approved):

Uluwatta:
Awesome!, thanks for the quick turnaround guys.

PR #1178, opened 2022-04-24
fix: add path type in acceptedFiles

aslan-alt:
- [x] Bug Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Style
- [ ] Build
- [ ] Chore
- [ ] Documentation
- [ ] CI

- [x] Not relevant

- [x] Not relevant

**Summary**
fix ts error

No

**Other information**

### DIFF
--- a/typings/react-dropzone.d.ts
+++ b/typings/react-dropzone.d.ts
@@ -1,6 +1,7 @@
 import * as React from "react";
 
-export { FileWithPath } from "file-selector";
+import { FileWithPath } from "file-selector";
+export { FileWithPath };
 export default function Dropzone(
   props: DropzoneProps & React.RefAttributes<DropzoneRef>
 ): JSX.Element;
@@ -23,7 +24,7 @@ export interface FileError {
 }
 
 export interface FileRejection {
-  file: File;
+  file: FileWithPath;
   errors: FileError[];
 }
 
@@ -68,7 +69,8 @@ export type DropzoneState = DropzoneRef & {
   isDragAccept: boolean;
   isDragReject: boolean;
   isFileDialogActive: boolean;
-  acceptedFiles: File[];
+  draggedFiles: FileWithPath[];
+  acceptedFiles: FileWithPath[];
   fileRejections: FileRejection[];
   rootRef: React.RefObject<HTMLElement>;
   inputRef: React.RefObject<HTMLInputElement>;


rolandjitsu:
@aslan-alt we cannot do this because the type of the files depends on what the `getFilesFromEvent` returns. So we use `T extends File` in places where we return file objects. I think you can set the `T` to the type you expect.

typings/react-dropzone.d.ts
```diff
@@ -66,8 +68,8 @@ export type DropzoneState = DropzoneRef & {
   isDragAccept: boolean;
   isDragReject: boolean;
   isFileDialogActive: boolean;
-  draggedFiles: File[];
-  acceptedFiles: File[];
+  draggedFiles: FileWithPath[];
+  acceptedFiles: FileWithPath[];
```
rolandjitsu, line 72: I've explained why you cannot do this. But we could use generics the same way we do for `onDrop`/`onDropAccepted`.

rolandjitsu (changes requested):

typings/react-dropzone.d.ts
aslan-alt: The variable T cannot be used because here is a concrete value and not a function
I changed the way of writing, hope you like it

typings/tests/hook.tsx
```diff
@@ -2,14 +2,6 @@ import React from "react";
 import { useDropzone, DropzoneProps } from "../../";
 
 export const Dropzone = ({ children, ...opts }: DropzoneProps) => {
-  const { ...state } = useDropzone(opts);
-  return children(state);
+  const state = useDropzone(opts);
+  return <>{children(state)}</>;
```
aslan-alt, line 6: Import React from "react" is used here; if there is no <> there will be the following error
 @typescript-eslint/no-unused-vars

typings/react-dropzone.d.ts
aslan-alt: Screen Shot 2022-04-25 at 14 37 59

typings/tests/hook.tsx
```diff
@@ -2,14 +2,6 @@ import React from "react";
...
 };
-
-<Dropzone>
-  {({ getRootProps, getInputProps }) => (
-    <div {...getRootProps()}>
-      <input {...getInputProps()} />
-    </div>
-  )}
-</Dropzone>;
```
rolandjitsu, line 15: Why was this removed?

typings/tests/hook.tsx
rolandjitsu: I haven't seen that error so far. How did you get it?

typings/tests/hook.tsx
gustavo-cardoso-deal: I'm curious too

aslan-alt: because this code is useless

aslan-alt: I don't think this code makes any sense
because there are no exports and references

aslan-alt: I don't think we've ever used these codes
because there are no exports and import
Screen Shot 2022-04-27 at 11 05 45

typings/tests/hook.tsx
aslan-alt: because I removed this useless code 🤣
Screen Shot 2022-04-27 at 11 20 04

typings/tests/hook.tsx
rolandjitsu: The purpose of it was to check that using the hook syntax doesn't cause any TS errors. So please add it back.

typings/tests/hook.tsx
rolandjitsu: Yeah, that code needs to be added back, then you won't get that warning.

aslan-alt: This is an optimization, I can delete it if you don't like it

typings/tests/hook.tsx
aslan-alt: This is an optimization, I can delete it if you don't like it

aslan-alt: ok, i have reverted the code
But I still want to know why not use the following way:
`
 return <>{children(state)}</>;
`
In this way, ts will not report errors, and the syntax is more streamlined

typings/tests/hook.tsx
aslan-alt: ok, i have reverted the code

typings/react-dropzone.d.ts
```diff
@@ -66,8 +66,8 @@ export type DropzoneState = DropzoneRef & {
   isDragAccept: boolean;
   isDragReject: boolean;
   isFileDialogActive: boolean;
-  draggedFiles: File[];
-  acceptedFiles: File[];
+  draggedFiles: (File & { path?: string })[];
+  acceptedFiles: (File & { path?: string })[];
```
rolandjitsu, line 70: Isn't `FileWithPath` the same thing?
```suggestion
  draggedFiles: FileWithPath[];
  acceptedFiles: FileWithPath[];
```
Could you try that and see if you still get errors?

aslan-alt: I always see this error, see [this demo](https://github.com/aslan-alt/react-dropzone-ts-demo)

rolandjitsu: @aslan-alt could you update to v14 and see if the same issue happens? I'll take a look later this week.

rolandjitsu: I just tried it out and if you set:
```
 acceptedFiles: FileWithPath[];
```
as suggested, the error is gone.

So please update that and also set the `FileRejection.file` prop type to the same type.

aslan-alt: okay

GiancarlosIO:
👀

aguscha333:
Any update on when this will be merged?

Crayon-ShinChan:
Will this PR be merged? I noticed that this project hasn't accepted any new PRs for the past two years. guys, what's going on?

rolandjitsu (approved):

PR #1301, opened 2023-05-08
feat: accept-folder-drop-overs

Codeprinz-1:
- [x] Bug Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Style
- [ ] Build
- [ ] Chore
- [ ] Documentation
- [ ] CI

- [x] Not relevant

- [ ] Not relevant

**Summary**
`react-dropdown` supports dropping folders but returns `isDragReject`: `true` when dragging folders. This fixes that

If this PR introduces a breaking change, please describe the impact and a migration path for existing applications.

**Other information**

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -54,7 +54,7 @@ export const TOO_MANY_FILES_REJECTION = {
 // that MIME type will always be accepted
 export function fileAccepted(file, accept) {
   const isAcceptable =
-    file.type === "application/x-moz-file" || accepts(file, accept);
+    file.type === "application/x-moz-file" || file.type === '' || accepts(file, accept);
   return [
     isAcceptable,
     isAcceptable ? null : getInvalidTypeRejectionErr(accept),


src/utils/index.js
```diff
@@ -54,7 +54,7 @@ export const TOO_MANY_FILES_REJECTION = {
 // that MIME type will always be accepted
 export function fileAccepted(file, accept) {
   const isAcceptable =
-    file.type === "application/x-moz-file" || accepts(file, accept);
+    file.type === "application/x-moz-file" || file.type === '' || accepts(file, accept);
```
rolandjitsu, line 57: I'm not too sure about this change and how it will affect the current behaviour. I'm trying to think if there are any other instances where the type may be missing but it's still a file and not a directory.

Also, the function is applied to files, not directories, so I'm a bit confused 🤔  Maybe you could provide some screencast of the issue you're trying to solve here.

rolandjitsu:
@Codeprinz-1 would you like to provide some example that illustrates the issue so that I can understand what this change is trying to achieve?

rolandjitsu:
Should be covered by `isDataTransferItemWithEmptyType` - https://github.com/react-dropzone/react-dropzone/blob/51e9af23b3d5ffb595fcba8b6b0952e2bdd805d7/src/utils/index.ts#L73

Codeprinz-1:
@rolandjitsu Not sure how I missed your earlier message to provide an example, but thanks for looking into this. I'll check out `isDataTransferItemWithEmptyType`

PR #1303, opened 2023-05-11
[Feature] Add ignoreNonFileDragEvents flag to prevent conflicts with other drag + drop libraries

iethree:
- [ ] Bug Fix
- [x] Feature
- [ ] Refactoring
- [ ] Style
- [ ] Build
- [ ] Chore
- [ ] Documentation
- [ ] CI

- [ ] Not relevant

- [ ] Not relevant

**Summary**
We are using `react-dropzone` as part of [a new feature in metabase ](https://github.com/metabase/metabase/issues/29561). Our application has a number of other drag-and-drop interactions throughout the application, and we found that we needed the ability to scope react-dropzone's event handling only to file drag events, because it interfered with other event types.

This pull request adds an optional `ignoreNonFileDragEvents` prop, which allows toggling of the intended behavior. Hopefully this option will be useful in other applications that may be relying on drag/drop events for other purposes.

No breaking changes. Simply adds a customization option.

**Other information**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -66,6 +66,7 @@ const defaultProps = {
   noKeyboard: false,
   noDrag: false,
   noDragEventsBubbling: false,
+  ignoreNonFileDragEvents: false,
   validator: null,
   useFsAccessApi: true,
   autoFocus: false,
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -872,7 +881,18 @@ export function useDropzone(props = {}) {
...
   const composeDragHandler = (fn) => {
-    return noDrag ? null : composeHandler(fn);
+    if (noDrag) {
+      return null;
+    }
+    if (ignoreNonFileDragEvents) {
+      return (event, ...args) => {
+        if (!isEvtWithFiles(event)) {
+          return;
+        }
```
iethree, line 891: Here's the heart of it: if the `ignoreNonFileDragEvents` flag is on and the event doesn't have files, we immediately return and let the event do whatever it wants.

iethree:
On further thought, this feature isn't that useful since you can externally compose the event handlers from `getRootProps()` 🤦‍♀️

PR #1385, opened 2024-10-14
fix: add file with path type for onDrop

jakeleventhal:
- [x] Bug Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Style
- [ ] Build
- [ ] Chore
- [ ] Documentation
- [ ] CI

- [x] Not relevant

- [x] Not relevant

**Summary**
The wrong type is used. Without this, manual typing is required.

No

**Other information**
N/A

### DIFF
--- a/typings/react-dropzone.d.ts
+++ b/typings/react-dropzone.d.ts
@@ -39,7 +39,7 @@ export type DropzoneOptions = Pick<React.HTMLProps<HTMLElement>, PropTypes> & {
   noDrag?: boolean;
   noDragEventsBubbling?: boolean;
   disabled?: boolean;
-  onDrop?: <T extends File>(
+  onDrop?: <T extends FileWithPath>(
     acceptedFiles: T[],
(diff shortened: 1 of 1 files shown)


typings/react-dropzone.d.ts
```diff
@@ -39,7 +39,7 @@ export type DropzoneOptions = Pick<React.HTMLProps<HTMLElement>, PropTypes> & {
   noDrag?: boolean;
   noDragEventsBubbling?: boolean;
   disabled?: boolean;
-  onDrop?: <T extends File>(
+  onDrop?: <T extends FileWithPath>(
```
rolandjitsu, line 42: `T` can be `FileWithPath` (you could either typecast or use `FileWithPath` as an arg to the cb function and typescript should pick up on it). It was intentional to keep the lower level type here instead of the `FileWithPath` because the type depends on what users use for the `getFilesFromEvent` func.

PR #1389, opened 2024-10-28
fix: make ESM build compatible with native Node.js

targos:
- Add extensions to relative import/export.
- Add `{"type": "module"}` to the ESM build directory.

- [x] Bug Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Style
- [ ] Build
- [ ] Chore
- [ ] Documentation
- [ ] CI

- [x] Not relevant

- [x] Not relevant

**Summary**
Explain the **motivation** for making this change. What existing problem does the pull request solve?
Try to link to an open issue for more information.

Allows to `import` the library from Node.js, for example with ESM-native SSR tools like Vite.

If this PR introduces a breaking change, please describe the impact and a migration path for existing applications.

It should not be breaking.

**Other information**

Fixes: https://github.com/react-dropzone/react-dropzone/issues/1259

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,9 +24,10 @@
   "scripts": {
     "cz": "git-cz",
     "clean": "rimraf ./dist",
-    "build": "yarn clean && yarn build:umd && yarn build:es",
+    "build": "yarn clean && yarn build:umd && yarn build:es && yarn build:es-package",
     "build:umd": "cross-env NODE_ENV=es rollup -c",
     "build:es": "cross-env BABEL_ENV=es babel ./src --out-dir ./dist/es --ignore '**/*.spec.js'",
+    "build:es-package": "echo '{\"type\":\"module\"}' > dist/es/package.json",
     "start": "styleguidist server",
     "styleguide": "styleguidist build",
     "test": "cross-env NODE_ENV=test yarn lint && jest --coverage && yarn typescript",
--- a/src/index.js
+++ b/src/index.js
@@ -26,7 +26,7 @@ import {
   onDocumentDragOver,
   pickerOptionsFromAccept,
   TOO_MANY_FILES_REJECTION,
-} from "./utils/index";
+} from "./utils/index.js";
 
 /**
  * Convenience wrapper component for the `useDropzone` hook
@@ -1039,4 +1039,4 @@ function reducer(state, action) {
 
 function noop() {}
 
-export { ErrorCode } from "./utils";
+export { ErrorCode } from "./utils/index.js";


package.json
```diff
@@ -24,9 +24,10 @@
   "scripts": {
     "cz": "git-cz",
     "clean": "rimraf ./dist",
-    "build": "yarn clean && yarn build:umd && yarn build:es",
+    "build": "yarn clean && yarn build:umd && yarn build:es && yarn build:es-package",
     "build:umd": "cross-env NODE_ENV=es rollup -c",
     "build:es": "cross-env BABEL_ENV=es babel ./src --out-dir ./dist/es --ignore '**/*.spec.js'",
+    "build:es-package": "echo '{\"type\":\"module\"}' > dist/es/package.json",
```
rolandjitsu, line 30: There's already a root package.json, why would this be necessary? There's some work in progress which should help with ESM: https://github.com/react-dropzone/react-dropzone/pull/1386.

rolandjitsu: @targos I wasn't aware that nesting a `package.json` the way you did will actually work the way you mention. Is this a bundler specific spec?

targos: It's not specific to bundlers. The spec comes from Node.js: https://nodejs.org/docs/latest/api/packages.html#type

targos: This PR fixes `import('react-dropzone')` in Node.js context.

rolandjitsu: Hmm, so according to:
> import statements of .js files are treated as ES modules if the nearest parent package.json contains "type": "module".

It would mean that the nested `package.json` specifying `"type": "module"` is what makes node interpret the dist/es as ESM.

Thanks for making it more clear as I wasn't aware that nested manifest files would result in such a behaviour.

akamfoad (approved):

szymonchudy:
Hey, hey! Are there plans to have it released soon? :)

TheRealFlyingCoder:
Bumping - this is a good change and it's been 4 months now 😄

cemalokten:
Looking forward to being able to use this in my react-19 project.

annesnour03:
Also need this change for my react 19 project, any way this could be merged?

rolandjitsu:
There's a comment which I was waiting on for a response. I'm concerned that if we start adding the `.js` extensions, it may start breaking things for many other users.

I did make an attempt at adding support for ESM in https://github.com/react-dropzone/react-dropzone/pull/1386 but got stuck on a few issues. And using a file extension started causing issues as far as I recall (though, I've used `.mjs` to denote an ES module).

targos:
I'm not aware of any tool that doesn't support explicit extensions in import specifiers but if you have a specific case that doesn't work I'm happy to look into it.

rolandjitsu:
> I'm not aware of any tool that doesn't support explicit extensions in import specifiers but if you have a specific case that doesn't work I'm happy to look into it.

I don't.

rolandjitsu (approved):

Issue #1400, opened 2024-11-09
[BUG] Not compatible with nextjs 15


Dakuan:
the latest version of nextjs is pinned with `react@"19.0.0-rc-02c0e824-20241028"` so attempting to install this lib doesnt work

```
➜  npm install react-dropzone
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: app@0.1.0
npm ERR! Found: react@19.0.0-rc-02c0e824-20241028
npm ERR! node_modules/react
npm ERR!   react@"19.0.0-rc-02c0e824-20241028" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer react@">= 16.8 || 18.0.0" from react-dropzone@14.3.5
npm ERR! node_modules/react-dropzone
npm ERR!   react-dropzone@"*" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR!
npm ERR!
npm ERR! For a full report see:
npm ERR! /Users/dominicbarker/.npm/_logs/2024-11-09T15_14_04_499Z-eresolve-report.txt```

**Comment by rolandjitsu** [comment]
I guess we'll have to test it. Btw, this is an enhancement since react 19 is still in RC.

**Comment by rolandjitsu** [comment]
A first look/try at using React 19 seems to pose some challenges: other dev dependencies we use are not supporting React 19 RC yet either.

It's interesting though that nextjs would default to React 19 which isn't even released as stable yet ...

**Comment by Philipp91** [comment]
React 19 has been released, so it would be great if react-dropzone worked with it.

FWIW, my `npm install` doesn't even complain, it just installs it alongwise react-dropzone 14.3.5. But then `tsc` has two little errors:

```
node_modules/react-dropzone/typings/react-dropzone.d.ts:7:4 - error TS2503: Cannot find namespace 'JSX'.

7 ): JSX.Element;
~~~

node_modules/react-dropzone/typings/react-dropzone.d.ts:11:36 - error TS2503: Cannot find namespace 'JSX'.

11   children?(state: DropzoneState): JSX.Element;
```

[...]

PR #1407, opened 2024-12-07
Use React.JSX instead of global JSX

Philipp91:
Related to #1400.

- [x] Bug Fix

- [x] Not relevant

- [x] Not relevant

**Summary**

To be compatible with React 19 typings, where global JSX was removed: https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript

I don't think it does. In particular, react-dropzone requires React v16.8 or newer, and there are `@types/react` releases for that old React version which include the new `React.JSX` type already. (I haven't verified this myself but read it online.)


Philipp91:
Withdrawing this in favor of #1410. But please do get that one merged.

PR #1459, opened 2026-06-04
fix: respect extensions for wildcard accept types

### LINKED ISSUE #1220: [BUG] `accept` doesn't work with wildcards
**Describe the bug**

Using a wildcard for `image/*` and then declaring file extensions doesn't seem to work as expected in Firefox and Brave/Chrome.

The following code will let users drop any image type and the jpeg/png file extensions are ignored.

**Note:** The same behaviour can be observed when the file picker dialog opens, it will allow picking other image extensions, not just jpeg and png.

```ts
useDropzone({
    accept: {
      'image/*': ['.jpeg', '.png']
    }
})
```

**To Reproduce**

1. Go to ['...'](https://react-dropzone.js.org/#section-accepting-specific-file-types)
2. Scroll to the 2nd example in the **Browser limitations** section (the one that "works").
3. Drag a webp, jpg, gif, heic, heif file onto it
5. See "all files will be accepted" message

**Expected behavior**
Should only accept `jpeg` and `png`.

The desired behaviour can be achieved by using multiple MIME types for each extension:
```ts
useDropzone({
    accept: {
      'image/png': ['.png'],
      'image/jpeg': ['.jpeg'],
   }
})
```

**Screenshots**
other-files

**Desktop (please complete the following information):**
 - OS: macOS 11.6.8
 - Browser: Brave
 - Version v1.41.100

puneetdixit200:
## Summary
- make wildcard MIME accept entries with extension lists reject files that match only the wildcard after drop/select
- omit wildcard MIME types from the hidden input accept attribute when extensions are provided, so the file picker can filter by extension
- add utility and dropzone regressions for `accept={{ "image/*": [".jpeg", ".png"] }}`

Fixes #1220

## Tests
- `yarn jest src/utils/index.spec.js src/index.spec.js --runInBand -t "wildcard MIME|sets only extensions"`
- `yarn jest src/utils/index.spec.js src/index.spec.js --runInBand`
- `yarn lint`
- `yarn test`
- `git diff --check`
- `yarn build && yarn size`

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -461,6 +461,13 @@ export function useDropzone(props = {}) {
   };
 
   const acceptAttr = useMemo(() => acceptPropAsAcceptAttr(accept), [accept]);
+  const inputAcceptAttr = useMemo(
+    () =>
+      acceptPropAsAcceptAttr(accept, {
+        omitWildcardMimeTypesWithExtensions: true,
+      }),
+    [accept]
+  );
   const pickerTypes = useMemo(() => pickerOptionsFromAccept(accept), [accept]);
 
   const onFileDialogOpenCb = useMemo(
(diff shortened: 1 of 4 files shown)


rolandjitsu:
Thanks for your contribution. A fix landed in https://github.com/react-dropzone/react-dropzone/releases/tag/v18.0.1.

PR #1489, opened 2026-07-19
feat: expose drag file rejections

### LINKED ISSUE #1288: [ENHANCEMENT] Expose fileRejections during dragging
**Is your feature request related to a problem? Please describe.**
Currently the reason why a file is being rejected during dragging is not accessible. After it is dropped, it's available through `fileRejections` but while dragging you only have the boolean variable named `isDragReject`.

To give proper feedback to the user it makes sense to expose the reasoning behind `isDragReject` so that you can return a message to the user about what is wrong with their file while they are dragging it.

**Describe the solution you'd like**
2 possible solutions.

1. Fill `fileRejections` while you are dragging and don't wait on the drop
2. Create a new variable `dragFileRejections` that is only filled when user is dragging

wanxiankai:
## Summary

- add `dragFileRejections` to the dropzone state so consumers can inspect reject reasons during drag
- compute drag-time rejection errors for type, size, and surplus-file checks without running the custom validator
- clear drag-time rejections on drag leave and drop so they stay separate from `fileRejections`

Fixes #1288.

## Tests

- `npm test -- --run src/index.spec.tsx`
- `npm run type-check`
- `npm run lint`
- `npm run format:check`

### DIFF
--- a/src/index.spec.tsx
+++ b/src/index.spec.tsx
@@ -2420,6 +2420,68 @@ describe("useDropzone() hook", () => {
       expect(dropzone).toHaveTextContent("dragReject");
     });
 
+    it("sets {dragFileRejections} with errors while files are dragged", async () => {
+      const {container} = render(
+        <Dropzone
+          accept={{
+            "image/*": []
+          }}
+        >
+          {({getRootProps, getInputProps, dragFileRejections, fileRejections}) => (
+            <div {...getRootProps()}>
+              <input {...getInputProps()} />
+              <span data-testid="drop-rejections">{fileRejections.length}</span>
+              <ul>
+                {dragFileRejections.flatMap(({errors}, rejectionIndex) =>
+                  errors.map(error => (
+                    <li key={`${rejectionIndex}-${error.code}`} data-type="drag-error">
+                      {error.code}
+                    </li>
+                  ))
+                )}
+              </ul>
+            </div>
+          )}
+        </Dropzone>
+      );
+      const dropzone = container.querySelector("div");
+
+      await act(() => fireEvent.dragEnter(dropzone, createDtWithFiles(files)));
+
+      expect(container.querySelector(`[data-testid="drop-rejections"]`)).toHaveTextContent("0");
+      expect(container.querySelectorAll(`[data-type="drag-error"]`)).toHaveLength(files.length);
+      expect(dropzone).toHaveTextContent("file-invalid-type");
+
+      fireEvent.dragLeave(dropzone, createDtWithFiles(files));
+
+      expect(container.querySelectorAll(`[data-type="drag-error"]`)).toHaveLength(0);
+    });
+
+    it("sets {dragFileRejections} for surplus files while dragging", async () => {
+      const {container} = render(
+        <Dropzone multiple={false}>
+          {({getRootProps, getInputProps, dragFileRejections}) => (
+            <div {...getRootProps()}>
+              <input {...getInputProps()} />
+              {dragFileRejections.map(({errors}, rejectionIndex) =>
+                errors.map(error => (
+                  <span key={`${rejectionIndex}-${error.code}`} data-type="drag-error">
+                    {error.code}
+                  </span>
+                ))
+              )}
+            </div>
+          )}
+        </Dropzone>
+      );
+      const dropzone = container.querySelector("div");
+
+      await act(() => fireEvent.dragEnter(dropzone, createDtWithFiles(images)));
+
+      expect(container.querySelectorAll(`[data-type="drag-error"]`)).toHaveLength(1);
+      expect(dropzone).toHaveTextContent("too-many-files");
+    });
+
     it("sets {isDragReject} if some files are too large", async () => {
       const {container} = render(
         <Dropzone maxSize={0}>
--- a/src/index.spec.tsx
+++ b/src/index.spec.tsx
@@ -2468,6 +2468,88 @@ describe("useDropzone() hook", () => {
       expect(dropzone).toHaveTextContent("dragReject");
     });
 
+    it("sets {dragFileRejections} with errors while files are dragged", async () => {
+      const {container} = render(
+        <Dropzone
+          accept={{
+            "image/*": []
+          }}
+        >
+          {({getRootProps, getInputProps, dragFileRejections, fileRejections}) => (
+            <div {...getRootProps()}>
+              <input {...getInputProps()} />
+              <span data-testid="drop-rejections">{fileRejections.length}</span>
+              <ul>
+                {dragFileRejections.flatMap(({errors}, rejectionIndex) =>
+                  errors.map(error => (
+                    <li key={`${rejectionIndex}-${error.code}`} data-type="drag-error">
+                      {error.code}
+                    </li>
+                  ))
+                )}
+              </ul>
+            </div>
+          )}
+        </Dropzone>
+      );
+      const dropzone = container.querySelector("div");
+
+      await act(() => fireEvent.dragEnter(dropzone, createDtWithFiles(files)));
+
+      expect(container.querySelector(`[data-testid="drop-rejections"]`)).toHaveTextContent("0");
+      expect(container.querySelectorAll(`[data-type="drag-error"]`)).toHaveLength(files.length);
+      expect(dropzone).toHaveTextContent("file-invalid-type");
+
+      fireEvent.dragLeave(dropzone, createDtWithFiles(files));
+
+      expect(container.querySelectorAll(`[data-type="drag-error"]`)).toHaveLength(0);
+    });
+
+    it("sets {dragFileRejections} for surplus files while dragging", async () => {
+      const {container} = render(
+        <Dropzone multiple={false}>
+          {({getRootProps, getInputProps, dragFileRejections}) => (
+            <div {...getRootProps()}>
+              <input {...getInputProps()} />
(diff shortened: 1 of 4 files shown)


src/index.tsx
```diff
@@ -449,6 +457,15 @@ export function useDropzone(props: DropzoneOptions = {}): DropzoneState {
               isDragReject: verdict === "reject",
               isDragUnknown: verdict === "unknown",
               isDragActive: true,
+              dragFileRejections: getDragFileRejections({
+                files: files as Array<FileWithPath | DataTransferItem>,
+                accept: inputAcceptAttr,
```
rolandjitsu: This makes `dragFileRejections` disagree with `isDragReject`. `getDragVerdict` (which drives `isDragReject`) is called with `acceptAttr`, but this uses `inputAcceptAttr` - and those differ on purpose: `inputAcceptAttr` drops a wildcard MIME when it's paired with extensions (it's for the `<input accept>` attribute, not match logic).

With `accept={{ "image/*": [".png"] }}`, dragging an image:
- `acceptAttr` = `image/*,.png` -> matches `image/*` -> verdict `accept`, `isDragReject = false`
- `inputAcceptAttr` = `.png` -> a dragged item has a MIME type but no filename, so `.png` can't match -> `dragFileRejections = [file-invalid-type]`

So a valid drag reports `isDragAccept === true` **and** a non-empty rejection list. Use the same source as the verdict:

```suggestion
                accept: acceptAttr,
```

Since `acceptAttr` is already in this callback's dependency array, the `inputAcceptAttr` you added to the deps ([L485](https://github.com/react-dropzone/react-dropzone/pull/1489/changes#diff-0b5adbfe7b36e4ae2f479291e20152e33e940f7f265162d77f40f6bdb5da7405R485)) can be dropped too.

src/index.tsx
```diff
@@ -995,4 +1017,54 @@ function reducer(state: DropzoneInternalState, action: any): DropzoneInternalSta
   }
 }
 
+function getDragFileRejections({
```
rolandjitsu: This duplicates the accept/size/surplus logic already in `getDragVerdict` (`src/utils/index.ts`) - and the two have already drifted (see the `acceptAttr` vs `inputAcceptAttr` note above). The drop path avoids this by computing per-file results once and deriving both the accepted list and the rejections from it (`setFiles`/`commit`).

I suggest making a single evaluator in `utils` return `{ verdict, rejections }` (or have `getDragVerdict` return the per-file results and derive the boolean), so `isDragReject` and `dragFileRejections` can't diverge and it stays unit-tested next to the existing `getDragVerdict` tests. It also removes the redundant second pass over the files on every drag event.

src/index.tsx
```diff
@@ -995,4 +1017,54 @@ function reducer(state: DropzoneInternalState, action: any): DropzoneInternalSta
...
+}): DragFileRejection[] {
+  const dragAcceptedFiles: Array<FileWithPath | DataTransferItem> = [];
+  const dragFileRejections: DragFileRejection[] = [];
+
+  const localizeError = (error: FileError, file: FileWithPath | DataTransferItem): FileError =>
+    getErrorMessage && file instanceof File ? {...error, message: getErrorMessage(error, file)} : error;
+
+  files.forEach(file => {
+    const [accepted, acceptError] = fileAccepted(file as File, accept);
+    const [sizeMatch, sizeError] = fileMatchSize(file as File, minSize, maxSize);
```
rolandjitsu: Worth a doc/JSDoc note: during a drag the items are `DataTransferItem`s exposing only MIME `type` - no name, no size (`getAsFile()` is null until drop). So `fileMatchSize` reads `undefined` and never fires, extension-based `accept` can't match, and the custom `validator` can't run. In practice `dragFileRejections` only ever surfaces `file-invalid-type` and `too-many-files` before drop. That's the expected browser limitation (same reason `getDragVerdict` returns `"unknown"` with a validator), but consumers should be told so they don't wait for size/validator reasons mid-drag.

src/index.spec.tsx
```diff
@@ -2420,6 +2420,68 @@ describe("useDropzone() hook", () => {
       expect(dropzone).toHaveTextContent("dragReject");
     });
 
+    it("sets {dragFileRejections} with errors while files are dragged", async () => {
+      const {container} = render(
+        <Dropzone
+          accept={{
+            "image/*": []
```
rolandjitsu, line 2475: This accept has no extensions, so `acceptAttr` and `inputAcceptAttr` are identical here - which hides the mismatch flagged on `index.tsx:462`. Please add a case with `accept={{ "image/*": [".png"] }}` dragging an `image/*` file and assert `isDragReject === false` **and** `dragFileRejections` is empty. That test fails on the current code and guards the fix.

rolandjitsu (changes requested):
Nice feature and clean API - the separate `dragFileRejections` field is the right call. One correctness blocker: it's computed from `inputAcceptAttr` while `isDragReject` uses `acceptAttr`, so the two can contradict each other (details inline, with a one-line fix + a test). The durable fix is to share one evaluator with `getDragVerdict` rather than re-implementing the checks. Happy to help once you've had a look.

src/index.tsx
wanxiankai: Fixed in `13d42b2`. The drag evaluator now uses `acceptAttr`, so wildcard MIME matching remains available during drag, and `inputAcceptAttr` is no longer a drag-callback dependency. I also added the requested integration regression for `accept={{ "image/*": [".png"] }}`, asserting `isDragReject === false` and an empty rejection list.

src/index.tsx
wanxiankai: Refactored in `13d42b2`: `evaluateDragFiles` in `src/utils/index.ts` now performs the type, size, and count checks once and returns both `{ verdict, rejections }`. `onDragEnter` derives all drag state from that single result, eliminating the duplicate pass and preventing drift. The shared evaluator also has a regression test next to the existing `getDragVerdict` tests.

src/index.tsx
wanxiankai: Added JSDoc on `DropzoneState.dragFileRejections` explaining that standard drag events expose `DataTransferItem`s with MIME type only, while extension, size, and custom-validator failures are available only after drop. It also states that standard browser drags can surface MIME-type and surplus-file errors here.

src/index.spec.tsx
wanxiankai: Added the requested integration regression in `src/index.spec.tsx` with `accept={{ "image/*": [".png"] }}` and an `image/gif` drag item. It asserts `isDragReject` is false and `dragFileRejections` is empty; the test fails on the previous implementation and passes with the shared evaluator.

wanxiankai:
Thanks for the detailed review — addressed in `13d42b2`:

- unified drag evaluation in `evaluateDragFiles`, returning both `verdict` and `rejections` from one pass and using `acceptAttr`
- added the wildcard MIME + extension integration regression and a utility-level regression
- documented the browser limitations of `dragFileRejections`

I also merged the latest `master` in `f537053` and resolved the conflict with the new grouped-accept support.

Local verification on the merged tree:

- `npm test -- --run` — 222 tests passed
- `npm run type-check`
- `npm run lint`
- `npm run format:check`

The two GitHub Actions workflows are currently awaiting maintainer approval.

src/utils/index.ts
```diff
@@ -207,6 +207,19 @@ export function allFilesAccepted({
  */
 export type DragVerdict = "accept" | "reject" | "unknown";
 
+/**
+ * A file rejected by the checks that can run during a drag.
+ */
+export interface DragFileRejection<T extends File | DataTransferItem = File | DataTransferItem> {
```
rolandjitsu, line 213: Can we consolidate this with the one you define in `utils/index.ts`? This would avoid drift b/t the 2. Unless you had a specific reason for keeping separate.

rolandjitsu (commented):
Thanks for the thorough turnaround. Your latest changes address mostly everything - with the exception of a small nit.

In principle, this LGTM.

src/utils/index.ts
wanxiankai: Good call, no specific reason to keep them separate. Addressed in the latest push by making the public `DragFileRejection` a type alias of the shared utils type: `UtilsDragFileRejection<FileWithPath | DataTransferItem>`. That keeps the state type shape in one place and avoids future drift.

I also re-ran the relevant local checks for this type-only change: `npm run type-check`, `npm run lint`, `npm run format:check`, and `npm test -- --run src/index.spec.tsx`.

rolandjitsu:
@wanxiankai the format needs fixing - CI shows failures.

wanxiankai:
Thanks for catching this — I fixed the remaining oxfmt issue in 3bd3a7f by formatting the type-only import in src/index.tsx.

Local verification on the current head:

- npm run format:check
- npm run lint
- npm run type-check
- npm test -- --run — 222 tests passed

The new Test workflow is currently awaiting maintainer approval. Could you please approve it and take another look when convenient?

rolandjitsu (approved):

rolandjitsu:
Please rebase and squash all commits into one.

wanxiankai:
Thanks — I rebased the branch onto the latest `master` and squashed the PR down to a single commit (`8cea790`).

Local verification on the rebased tree:
- `npm run format:check`
- `npm run lint`
- `npm run type-check`
- `npm test -- --run` — 222 tests passed

The GitHub checks have been retriggered; two workflows are currently awaiting maintainer approval.

